PR #41, opened 2020-08-11
Added F marco to serial print strings

Mikhail-Beresnev:
Added F(...) in every Serial, Serial1, and client .print and .println statement, resulting in .print(F("...")) and .println(F("...")) statements
Reduced dynamic memory global variable usage from 5074 bytes (out of 8192, 61%) to 3578 bytes (out of 8192, 43%)


Open_Acidification_pH-stat_arduino/printSpecifiedLines.ino
```diff
@@ -1,20 +1,20 @@
...
+    Serial.print(F("NUMLINES: "));
     Serial.println(numLines);
     int printedLines = 0;
     // seek starting byte
     long startingByte = startingLine * (RECORD_LENGTH + 2);
     if (myFile.seek(startingByte)) {
-      Serial.println("SUCCESSFUL SEEK");
+      Serial.println(F("SUCCESSFUL SEEK"));
       printHeader(client, 200);
       client.println("time,tankid,temp,temp setpoint,pH,pH setpoint,onTime");
```
prestoncarman, line 14: I think you can also use F macro for `client.println` and `client.print`. Can you try it and see if it compiles?

prestoncarman:
Also please fix the clang format error, so all tests are passing.

PR #43, opened 2020-08-23
Move specific global variables to local files

Mikhail-Beresnev:
Moved global variables that were only used in one file to that particular file. Removed variables which are declared but are never used.


Open_Acidification_pH-stat_arduino/Open_Acidification_pH-stat_arduino.ino
```diff
@@ -154,7 +135,6 @@ const int TEMP_CORR_ADDRESS = 12;
 const int KP_ADDRESS = 20;
 const int KI_ADDRESS = 28;
 const int KD_ADDRESS = 36;
-const int MAC_ADDRESS = 44;
```
prestoncarman, line 157: I think I would leave this since the MAC_ADDRESS is still being stored at this address. Think of this as part of the documentation.

PR #52, opened 2020-10-15
Use singleton for TankControllerLib

jgfoster:
This should be coordinated with https://github.com/Open-Acidification/TankControllerLib/pull/3. 

Also, remove include of Wire.h since it is not used.


prestoncarman (approved):
LGTM

PR #56, opened 2020-11-02
WIP: A bit of Chiller encapsulation

lsaca05:
Not sure how to test.

Also need to reconsider CO2 encapsulation.


lsaca05:
Didn't realize there was a TankControllerLib repository. This would be more appropriate over there. Closing.

PR #70, opened 2021-03-29
PHProbe wrapper part 2

eucalvo:
Added Temperature Compensation


src/Devices/PHProbe.cpp
```diff
@@ -28,6 +28,13 @@ PHProbe::PHProbe() {
   Serial1.print(F("C,1\r"));    // Reset pH stamp to continuous measurement: once per second
 }
 
+void PHProbe::onePointCalibration(double midpoint) {
+  const String PARTIAL_COMMAND = "Cal,mid,";
```
jgfoster, line 32: We had a partial command in the earlier function (temperature?) because we had branching such that the same literal was used twice. Here we don't have branching so there is less reason to make an extra variable. We might just as well add it to the expression below as a literal.

src/Devices/PHProbe.cpp
```diff
@@ -41,3 +48,28 @@ void PHProbe::serialEvent1() {
     serial->println();
   }
 }
+
+// "pH decreases with increase in temperature. But this does not mean that
+//  water becomes more acidic at higher temperatures."
+// https://www.westlab.com/blog/2017/11/15/how-does-temperature-affect-ph
+void PHProbe::setTemperatureCompensation(double temperature) {
```
jgfoster, line 55: Why does this show as a diff? Did you (fail to) update your branch from main?

src/Devices/PHProbe.cpp
```diff
@@ -41,3 +48,28 @@ void PHProbe::serialEvent1() {
...
+    fullCommand = PARTIAL_COMMAND + String(temperature, 2) + "\r";
+  } else {
+    fullCommand = PARTIAL_COMMAND + "20\r";
+  }
+  Serial.println(fullCommand);
+  Serial1.print(fullCommand);  // send that string to the Atlas Scientific product
+}
+
+void PHProbe::twoPointCalibration(double lowpoint, double midpoint) {
+  const String MIDPOINT_PARTIAL_COMMAND = "Cal,mid,";
```
jgfoster, line 68: As above, I don't think we need separate strings for the literals.

src/Devices/PHProbe.h
```diff
@@ -1,13 +1,23 @@
 #pragma once
 #include "Arduino.h"
 
+/**
+ * Issuing the cal,mid command after the
+ * EZO pH circuit has been calibrated, will
+ * clear the other calibration points.
+ * Full calibration will have to be redone
```
jgfoster, line 8: Is a "full calibration" the same as `twoPointCalibration()`? If so, we should say so. If not, how do we do a full calibration?

test/PHProbeTest.cpp
```diff
@@ -29,4 +29,37 @@ unittest(serialEvent1) {
   assertEqual(7.75, PHProbe::instance()->getPH());
 }
 
+unittest(setTemperatureCompensation) {
```
jgfoster, line 32: I believe this was already merged.

jgfoster (changes requested):
You've done well at applying what we did with the temperature to the additional functions. Nice work!

src/Devices/PHProbe.cpp
```diff
@@ -28,16 +29,83 @@ PHProbe::PHProbe() {
   Serial1.print(F("C,1\r"));    // Reset pH stamp to continuous measurement: once per second
 }
 
+void PHProbe::clearCalibration() {
+  Serial1.print("Cal,clear\r");  // send that string to the Atlas Scientific product
+}
+
+String PHProbe::getSlope() {
```
jgfoster, line 36: Do you have some pending changes to this function? I thought we set up a `sendReadSlope()` function that did not block. This function blocks when it calls serialEvent1().

src/Devices/PHProbe.cpp
```diff
@@ -28,16 +29,83 @@ PHProbe::PHProbe() {
...
+          slopeResponse = string;
+        }
+      }
+    }
+  }
+}
+
+double PHProbe::getPhReading() {
+  TankControllerLib *pTC = TankControllerLib::instance();
+  pTC->serialEvent1();
```
jgfoster, line 81: This will block and I don't think we want that.

jgfoster (changes requested):
Let's discuss this...

src/Devices/PHProbe.cpp
```diff
@@ -1,6 +1,7 @@
 #include "Devices/PHProbe.h"
 
 #include "Devices/Serial_TC.h"
+#include "TankControllerLib.h"
```
jgfoster, line 4: What code relies on this?

jgfoster (approved):
Looks about ready.

prestoncarman (approved):
LGTM

jgfoster (approved):

PR #74, opened 2021-04-02
Update GUI with fewer boxes and more density-independent layout

jgfoster:
Let's see how this looks on Ubuntu.


prestoncarman (approved):
LGTM

PR #76, opened 2021-04-07
See PID constants

Kavikick:


test/SeePIDConstants.cpp
```diff
@@ -0,0 +1,61 @@
...
+  LiquidCrystal_TC* display = LiquidCrystal_TC::instance();
+  PID* pPID = PID_TC::instance()->getPID();
+  SeePIDConstants* test = new SeePIDConstants(tc);
+  char lines[2][17];
+
+  // Set up
+  double kp = pPID->GetKp();
+  double ki = pPID->GetKi();
+  double kd = pPID->GetKd();
+  pPID->SetTunings(10000.0, 10000.0, 10000.0);
```
jgfoster, line 23: We should probably have different values so we can confirm that the correct values are being read.

src/UIState/SeePIDConstants.cpp
```diff
@@ -4,12 +4,72 @@
...
+  if (msEnd < millis()) {
+    COUT("done");
+    returnToMainMenu();
+  }
+}
+
+void SeePIDConstants::loadKp(int line) {
+  PID *pPID = PID_TC::instance()->getPID();
+  char buffer[17];
+  sprintf(buffer, "kp:%.1f", pPID->GetKp());
```
jgfoster, line 53: It appears that the convention is to use an uppercase 'K' for these constants. Also, depending on the number of digits we expect to need, it might be nice to have a space between the label (including the colon) and the data.

src/UIState/SeePIDConstants.cpp
```diff
@@ -4,12 +4,72 @@
...
+  LiquidCrystal_TC::instance()->writeLine(buffer, line);
```
jgfoster, line 54: I suspect that the LCD does not clear the rest of the line. So far we've been making sure that everything we send to the LCD is exactly 16 characters long and this seems like it is putting the effort in the wrong place. I think that we should modify LiquidCrystal_TC::writeLine() so that it either clears the line first or adds spaces to the end of the string to be printed (copying into an adequate buffer first.

jgfoster (changes requested):

src/UIState/SeePIDConstants.cpp
jgfoster: See #80.

src/Devices/LiquidCrystal_TC.cpp
```diff
@@ -86,5 +88,19 @@ void LiquidCrystal_TC::writeLine(const char* text, int line) {
   setCursor(0, line);
   print("                ");
```
jgfoster, line 87: When I wrote up the issue I didn't realize that we already have code that clears the entire line (my bad!). It seems to me that either the line above isn't needed or the code below isn't needed. Thoughts?

src/UIState/SeePIDConstants.cpp
```diff
@@ -4,12 +4,68 @@
...
+#include "Devices/PHProbe.h"
+#include "Devices/PID_TC.h"
+#include "Serial_TC.h"
+#include "TC_util.h"
+
+void SeePIDConstants::loop() {
+  int elapsedSeconds = (millis() - startTime) / 1000;
+
+  Serial_TC *mySerial = Serial_TC::instance();
+  mySerial->print("SeePIDConstants::loop() - 1 CurrentTime ", elapsedSeconds);
```
jgfoster, line 17: If this was for debugging, it should come out. I don't think we want to log a new line every time through the loop.

src/UIState/SeePIDConstants.cpp
```diff
@@ -4,12 +4,68 @@
...
+      loadSlope(0);
+      loadKp(1);
       break;
-  };
+    case 4:
+      loadKp(0);
+      loadKi(1);
+      break;
+  }
+  if (elapsedSeconds >= 5) {
```
jgfoster, line 40: Could this be a default case in the switch statement?

jgfoster (changes requested):
Just a couple more things...

src/UIState/SeePIDConstants.cpp
Kavikick: With the way things are currently, having it in it's own if gives no benefit, so you're right a default would do the same thing.

src/Devices/LiquidCrystal_TC.cpp
Kavikick: I think the code below should stay because it has the capacity to trim larger strings down to 16 characters

jgfoster: agreed

jgfoster (approved):
LGTM

prestoncarman (approved):
LGTM

prestoncarman:
Oh, something failed.

jgfoster (approved):
LGTM

PR #85, opened 2021-04-14
KP, KI, KD setters

eucalvo:


src/UIState/MainMenu.h
```diff
@@ -32,7 +32,9 @@ class MainMenu : public UIState {
     SET_CALIBRATION_CLEAR,
     SET_TEMP_CALIBRATION,
     SET_PID_AUTO_TUNE,
-    SET_PID_MANUAL_TUNE,
+    SET_KP,
```
jgfoster, line 35: alphabetize these

src/UIState/SetKD.cpp
```diff
@@ -0,0 +1,17 @@
+/**
+ * SetKD.cpp
+ */
+
+#include "SetKD.h"
+
+#include "../Devices/EEPROM_TC.h"
```
jgfoster, line 7: See if it compiles without the "../" (see, e.g., scr/UIState/MainMenu.cpp:4).

jgfoster (changes requested):
This is nice. I've requested a couple small changes.

prestoncarman (approved):
LGTM after you fixed @jgfoster comments.

src/UIState/MainMenu.h
jgfoster: Actually, no, we want the menus to show in a priority or "natural" order.

PR #88, opened 2021-04-15
Fix time display for GUI

### LINKED ISSUE #84: View uptime crashes in Python GUI

jgfoster:
Fix #84. Calculation of offset between host and Arduino mock was wrong and has been fixed. Also, we don't need the `floor()` function (which returns a float) when we are doing integer division.


prestoncarman (approved):
LGTM

PR #92, opened 2021-04-19
A Collection of Views

Kavikick:
Fixed the naming of the instance in Ethernet_TC
Created views for
- 	SeeDeviceAddress
- 	SeeGoogleMins
- 	SeePHSlope(Pulled out of SeePIDConstants)
- 	SeeTankID
- 	SeeVersion

P.S: Waiting on #91 to add view for SeeLogFile


src/UIState/SeeDeviceAddress.cpp
```diff
@@ -4,12 +4,17 @@
...
 
-void SeeDeviceAddress::handleKey(char key) {
-  switch (key) {
-    case 'D':  // Don't save (cancel)
-      returnToMainMenu();
-      break;
-    default:
-      break;
-  };
+#include "../Devices/Ethernet_TC.h"
```
jgfoster, line 7: Is the "../" necessary?

src/UIState/SeeDeviceAddress.cpp
```diff
@@ -4,12 +4,17 @@
...
+#include "../Devices/LiquidCrystal_TC.h"
+#include "Ethernet.h"
+
+void SeeDeviceAddress::loop() {
+  IPAddress IP = Ethernet_TC::instance()->getIP();
+  byte* mac = Ethernet_TC::instance()->getMac();
+  char line0[170];
```
jgfoster, line 14: `170`?

src/UIState/SeeDeviceAddress.cpp
```diff
@@ -4,12 +4,17 @@
...
-      returnToMainMenu();
-      break;
-    default:
-      break;
-  };
+#include "../Devices/Ethernet_TC.h"
+#include "../Devices/LiquidCrystal_TC.h"
+#include "Ethernet.h"
+
+void SeeDeviceAddress::loop() {
```
jgfoster, line 11: Is there a reason to use `loop()` instead of `start()`? It seems that if it doesn't change, then there isn't any need to refresh it many times a second.

src/UIState/SeeGooglemins.cpp
```diff
@@ -0,0 +1,14 @@
+/**
+ * SeeGoogleMins.cpp
```
jgfoster, line 2: The file name should have a capital 'M'.

src/UIState/SeeGooglemins.cpp
```diff
@@ -0,0 +1,14 @@
...
+ */
+
+#include "SeeGoogleMins.h"
+
+#include "../Devices/EEPROM_TC.h"
+#include "../Devices/LiquidCrystal_TC.h"
+
+void SeeGoogleMins::loop() {
+  char buffer[17];
+  sprintf(buffer, "%i", EEPROM_TC::instance()->getGoogleSheetInterval() / 60000);
```
jgfoster, line 12: Are we storing minutes or milliseconds? What happens in `SetGoogleSheetInterval.cpp`?

jgfoster (changes requested):

src/UIState/SeeDeviceAddress.cpp
```diff
@@ -4,12 +4,17 @@
...
 
-void SeeDeviceAddress::handleKey(char key) {
-  switch (key) {
-    case 'D':  // Don't save (cancel)
-      returnToMainMenu();
-      break;
-    default:
-      break;
-  };
+#include "../Devices/LiquidCrystal_TC.h"
```
jgfoster, line 7: Do we need the "../" as part of the path?

src/UIState/SeeGoogleMins.cpp
```diff
@@ -0,0 +1,14 @@
+/**
+ * SeeGoogleMins.cpp
+ */
+
+#include "SeeGoogleMins.h"
+
+#include "../Devices/EEPROM_TC.h"
```
jgfoster, line 7: Do we need "../" as part of the path?

src/UIState/SeeGoogleMins.cpp
```diff
@@ -0,0 +1,14 @@
...
+#include "../Devices/LiquidCrystal_TC.h"
+
+void SeeGoogleMins::loop() {
```
jgfoster, line 10: Is there a reason to use `loop()` rather than `start()`?

jgfoster (changes requested):

src/UIState/SeeGooglemins.cpp
Kavikick: Looking at 0.197 we collect the value in minutes.

src/UIState/SeeVersion.cpp
```diff
@@ -0,0 +1,14 @@
+/**
+ * SeeVersion.cpp
+ */
+
+#include "SeeVersion.h"
+
+#include "../TankControllerLib.h"
```
jgfoster, line 7: Does it work without the "../" prefix?

jgfoster (changes requested):

src/UIState/SeeVersion.cpp
Kavikick: Yup, it does work!

Kavikick: At least locally

Kavikick: I miss read that, my bad. I'll check that really quick.

Kavikick: It works

jgfoster (approved):
LGTM

src/UIState/SeePHSlope.cpp
```diff
@@ -0,0 +1,12 @@
...
+ * SeePHSlope.cpp
+ */
+
+#include "SeePHSlope.h"
+
+#include "Devices/LiquidCrystal_TC.h"
+#include "Devices/PHProbe.h"
+
+void SeePHSlope::loop() {
+  LiquidCrystal_TC::instance()->writeLine(PHProbe::instance()->getSlope().c_str(), 1);
```
prestoncarman, line 11: Does the slope continually change?

test/SeePIDConstants.cpp
```diff
@@ -31,7 +27,7 @@ unittest(TestVerticalScroll) {
   tc->loop();
   assertEqual("SeePIDConstants", tc->stateName());
 
-  // during the delay we cycle through kp,ki,kd, and ph slope
+  // during the delay we cycle through kp,ki, and kd
   assertEqual("Kp: 10001.0     ", display->getLines().at(0));
```
prestoncarman, line 31: I thought the view was going to just show kp and ki, then show kd, then go back to kp and ki.

src/UIState/SeePIDConstants.cpp
```diff
@@ -10,7 +10,7 @@
 #include "TC_util.h"
 
 void SeePIDConstants::loop() {
-  int elapsedSeconds = (millis() - startTime) / 1000;
+  int elapsedSeconds = ((millis() - startTime) / 1000) % 3;
   switch (elapsedSeconds) {
     case 0:
       loadKp(0);
```
prestoncarman, line 16: What happened to the blank line?

src/UIState/SeePHSlope.cpp
jgfoster: I agree that this should transfer over to `start()`. Perhaps do a search for `loop()` to see if we missed any others.

src/UIState/SeePIDConstants.cpp
jgfoster: What we showed today was three seconds per "frame".

test/SeePIDConstants.cpp
jgfoster: What we discussed was a loop of the following:
"Kp / Ki"  then "Ki / Kd" then "Kd / blank"
An alternate would be "Kp / Ki" then "Kd / blank" (no need to give Ki double listing).

src/UIState/SeePHSlope.cpp
Kavikick: Currently no. I'll change it to start()

src/UIState/SeePIDConstants.cpp
```diff
@@ -10,31 +10,15 @@
 #include "TC_util.h"
 
 void SeePIDConstants::loop() {
-  int elapsedSeconds = (millis() - startTime) / 1000;
+  int elapsedSeconds = ((millis() - startTime) / 1000) % 2;
```
jgfoster, line 13: I think we can edit this to have a 3-second delay.

jgfoster (approved):

PR #94, opened 2021-04-21
Changed all sprintf To snprintf

EvanKropf:
in reference to https://github.com/Open-Acidification/TankController/issues/65

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -109,7 +109,7 @@ RTDINLOW
 setpoint
 shutil
 sizeof
-sprintf
+snprintf
 src
 strcpy
 strlen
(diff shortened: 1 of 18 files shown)


jgfoster:
The spell checking test has the following output:

```
New misspellings found, please review:\n\n* smprintf\n* snprintf\
```

This tells us that there are two unrecognized words: `smprintf` and `snprintf`. The first one looks like a typo so you should fix it. The second one is legitimate, so should be added as a valid word. This will be done in `.github/actions/spelling/expect.txt` after line 111. Note that we already have `sprintf` as an allowed word and I'd like you to remove that word so that any future use of the old function will be caught.

jgfoster (approved):
That looks like just what I asked for!

prestoncarman (approved):
LGTM

prestoncarman:
@all-contributors please add @EvanKropf for code

prestoncarman:
@all-contributors please add @EvanKropf for code

PR #95, opened 2021-04-22
Ph calibration: midpoint, lowpoint, highpoint

eucalvo:


src/UIState/MainMenu.cpp
```diff
@@ -170,15 +170,12 @@ void MainMenu::selectView() {
...
-      break;
-    case SET_CALIBRATION_2:
-      this->setNextState((UIState *)new PHCalibration(tc));
-      break;
-    case SET_CALIBRATION_CLEAR:
-      this->setNextState((UIState *)new PHCalibration(tc));
+    case SET_CALIBRATION:
+      this->setNextState((UIState *)new PHCalibrationMid(tc));
       break;
+    // case SET_CALIBRATION_CLEAR:
```
jgfoster, line 176: Since this is part of MainMenu we should probably implement at least a stub that does nothing.

jgfoster (changes requested):
Just a couple lines of commented-out code (as well as whatever spelling or other tests remain).

jgfoster:
Rather than using `sprint()` we should use `snprintf()`. See other examples in the code.

src/UIState/MainMenu.cpp
jgfoster: See #96.

jgfoster (approved):
I've addressed the issues I raised.

prestoncarman (approved):
LGTM

PR #97, opened 2021-04-26
Refactor pH Control

jgfoster:
* Update pH control each loop
* Refactor PHControl
  * use a singleton
  * save and retrieve target in EEPROM
  * add some debugging
* MainMenu idle loop now shows actual pH target
* SetPHSetPoint UI State now assigns to PHControl singleton instead of EEPROM
* Update tests to match new design


prestoncarman (approved):
LGTM

PR #104, opened 2021-04-27
GUI updates for temp & CO2

jgfoster:
* Show digital pins for heat/chill and CO2 bubbler in GUI
* Log temp and CO2 pin values when they change
* Trim CR from serial 1 input before logging


prestoncarman (approved):
LGTM

PR #107, opened 2021-04-27
Stop control when doing calibration (fix #102)

### LINKED ISSUE #102: Stop control when doing calibration
For both temp and pH.

jgfoster:
Also, don't return to idle menu after 60 seconds.


prestoncarman (approved):
LGTM

Created #108 to highlight issue introduced by this change.

PR #120, opened 2021-04-30
Remove directory path from include files

jgfoster:
Seems that it is not needed!


jgfoster:
It seems that the path isn't needed for the arduino_ci tests, but it is needed to actually compile to the Arduino.

PR #128, opened 2021-05-03
Clear calibration with 'A'

jgfoster:
No need to use NumberCollectorState; just use 'A' for accept and 'D' for return to main menu.


prestoncarman (approved):
LGTM

PR #134, opened 2021-05-06
See ph slope

eucalvo:
and template class for PID on/off


src/UIState/MainMenu.cpp
```diff
@@ -203,7 +199,7 @@ void MainMenu::selectSet() {
       this->setNextState((UIState *)new SetKP(tc));
       break;
     case SET_PID_ON_OFF:
-      this->setNextState((UIState *)new PIDTuningMenu(tc));
+      this->setNextState((UIState *)new SetPIDOnOff(tc));
```
jgfoster, line 202: We have a new code style checker that is stricter about some things that we have been in the past. This line should be revised as follows:
`   this->setNextState(static_cast<UIState *>(new SetPIDOnOff(tc)));`
Eventually all the other "C-style type casts" should be converted to "C++-style type casts" but for now I'm just doing them one-at-a-time as the tests complain.

src/UIState/SeePHSlope.cpp
```diff
@@ -7,7 +7,12 @@
...
 
-void SeePHSlope::start() {
-  LiquidCrystal_TC::instance()->writeLine(prompt(), 0);
+void SeePHSlope::loop() {
   LiquidCrystal_TC::instance()->writeLine(PHProbe::instance()->getSlope().c_str(), 1);
 }
+
+void SeePHSlope::start() {
+  LiquidCrystal_TC::instance()->writeLine(prompt(), 0);
+  LiquidCrystal_TC::instance()->writeLine("requesting slope", 1);
```
jgfoster, line 16: See #135.

jgfoster (changes requested):
Good work! I have a couple small suggestion...

prestoncarman (approved):
LGTM

PR #136, opened 2021-05-06
Use flash memory for most strings

jgfoster:
Reduce text in format string and put most string literals as arguments.


jgfoster:
Strings in flash doesn't play nicely with `sprintf()` so we will take them out for now.

PR #138, opened 2021-05-06
Flash doesn't play nicely with `sprintf()`

jgfoster:
We will remove them for now.


prestoncarman (approved):
LGTM

Lets get it working then optimize.

PR #140, opened 2021-05-07
Changes so that float values are printed

jgfoster:
Arduino's default C library has an implementation of printf() that does not support float. We work around this by printing two integers.


jgfoster:
Withdrawing this for now...

PR #148, opened 2021-05-12
Get things working on the board

jgfoster:
* Remove `serialWithTime()` and add `DateTime_TC::printToSerial()`
* Use `Serial.println()` instead of `serial()` during startup
* Changes to pH Probe and Control based on on-device testing
* Temperature should only be read once per second
* Keep all files at root and do not use directories
* All file names should be 8.3 size
* Update some debugging output


prestoncarman (approved):
LGTM

Issue #146, opened 2021-05-10
Remove milliseconds from SerialWithTime 


prestoncarman:
It does not make sense to append uptime milliseconds to RTC datetime. We need to display milliseconds in a different way. Lets discuss alternate options:

* Show device uptime
* Show delta since last called
* ...

jgfoster:
After some discussions with Eliam, it occurs to me that the reason I added the milliseconds was so you could do the mental arithmetic of calculating how long it had been since the solenoid had changed. Instead let's just tell the user how long it has been! So we just say "heater off after 400 ms". 

With this I'd like to retire `serialWithTime()` and change Serial_TC::vprintf() so that it keeps track of the last serial event time and if it has been more that a minute (for example) it adds a timestamp.

jgfoster:
Fixed #150.

jgfoster:
Done!

PR #150, opened 2021-05-12
Serial out temp & pH

eucalvo:


jgfoster (changes requested):
It looks like we have a couple failing tests. You should pull from main to get the latest so you can observe the same test failures. To get rid of the RTC warning you could clear the output buffer just before running the test. Then, you need to add the date/time line to your expected output.

jgfoster (approved):
LGTM

jgfoster:
Fixed #146.

test/TestTemperatureControl.cpp
```diff
@@ -60,19 +62,23 @@ unittest(AfterIntervalAndWithinDelta) {
...
+  DateTime_TC january(2021, 1, 15, 1, 48, 24);
+  january.setAsCurrent();
+  state->serialPort[0].dataOut = "";  // the history of data written
   // chiller is initially off and goes on when needed
   assertEqual(TURN_SOLENOID_OFF, state->digitalPin[PIN]);
   chiller.setTargetTemperature(20);
   delay(31000);
   chiller.updateControl(20.05);
   assertEqual(TURN_SOLENOID_ON, state->digitalPin[PIN]);
+  assertEqual("2021-01-15 01:48:55\r\nchiller on after 31000 ms\r\n", state->serialPort[0].dataOut);
```
prestoncarman, line 81: We should also have a test print for when the chiller turns off.

test/TestTemperatureControl.cpp
```diff
@@ -116,11 +122,13 @@ unittest(OutsideDelta) {
   GodmodeState* state = GODMODE();
   state->reset();
   Heater heater;
+  state->serialPort[0].dataOut = "";  // the history of data written
   // heater is initially off, then turns on
   assertEqual(TURN_SOLENOID_OFF, state->digitalPin[PIN]);
   heater.setTargetTemperature(20);
   heater.updateControl(19.95);
   assertEqual(TURN_SOLENOID_ON, state->digitalPin[PIN]);
+  assertEqual("2021-01-15 01:48:24\r\nheater on after 0 ms\r\n", state->serialPort[0].dataOut);
```
prestoncarman, line 131: We should also have a test print for when the heater turns off.

src/Devices/PHControl.h
```diff
@@ -5,6 +5,7 @@ class PHControl {
   // Class variable
   static PHControl* _instance;
   // instance variables
+  unsigned long lastSwitchMS = 0;
```
prestoncarman, line 8: This seems out of place since you are only changing temperature.

Remove this line and include it in the ph change.

prestoncarman (changes requested):

prestoncarman (approved):
LGTM

jgfoster (approved):

PR #155, opened 2021-05-18
Use `float` instead of `decimal`

jgfoster:
The Arduino Mega2560 uses 32 bits for `double`, the same as `float`. It will be easier to understand the code if we use the more common type.


prestoncarman (approved):
LGTM

PR #157, opened 2021-05-18
Revert to a single EEPROM wrapper

jgfoster:
Since the `double` is only 4 bytes, we don't have any overlap in the memory.


prestoncarman:
Should this be using double or float? I think we are switching to float.

prestoncarman (approved):
LGTM

Lets keep it simple

PR #159, opened 2021-05-20
Replace `int` with `int16_t` or variations

jgfoster:
Different compilers and platforms have different sizes for integers. It is a best practice to be explicit as to the sizes.


prestoncarman (approved):
LGTM

PR #160, opened 2021-05-20
Python GUI changes

jgfoster:
* Use python3 explicitly for GUI demo
* Fix problem with opening SD from GUI demo
* Rename `build` to `.build`


prestoncarman (approved):
LGTM

PR #161, opened 2021-05-20
Number collector changes

jgfoster:
* Show new value message after saving entered number
* Show number as 32-bit integer rather than as 16-bit integer


prestoncarman (approved):
LGTM

PR #163, opened 2021-05-24
PushingBox

eucalvo:


src/Devices/Ethernet_TC.h
```diff
@@ -5,7 +5,12 @@
 
 #pragma once
 
-#include "Ethernet.h"
+#include <Arduino.h>
+#ifdef MOCK_PINS_COUNT
+#include <Ethernet_CI.h>
```
jgfoster, line 10: The file <Ethernet_CI.h> begins with the following:

```
#include <Ethernet.h>
#ifdef MOCK_PINS_COUNT
```

This means that you can include it unconditionally and it will include <Ethernet.h> for you.

src/Devices/PushingBox.cpp
```diff
@@ -0,0 +1,65 @@
...
+    sendData();
+    unsigned long interval = EEPROM_TC::instance()->getGoogleSheetInterval();
+    // jump to the next multiple of interval
+    nextSendTime = (now / interval + 1) * interval;
+  }
+  // are we still connected?
+  if (client.connected()) {
+    // if so, read response
+    int next;
+    while ((next = client.read()) != -1) {
```
jgfoster, line 37: Codacy "is a very simple pattern-matching tool, and all it cares about there is that the read function is being used - it doesn't parse the arguments or do any semantic analysis. It's not really intended for use without a human looking over its shoulder." [cite](https://github.com/jamulussoftware/jamulus/issues/1081). I'm not sure how we should handle this. Perhaps there is a way to suppress the warning for a particular line. Or, we could just accept the PR and Codacy won't complain again until the line is edited.

jgfoster (approved):
It seems that the failing tests are related to timing (see Blink.cpp). Did we add some `delay()` calls that occur in the general case? Otherwise, this looks good.

.github/actions/spelling/expect.txt
```diff
@@ -1,8 +1,10 @@
 Adafruit
 admins
+api
 arduino
 args
 ASingle
+attepting
```
prestoncarman, line 7: I don't think this is spelled correctly.

prestoncarman (changes requested):

src/Devices/PushingBox.h
```diff
@@ -0,0 +1,32 @@
...
+
+private:
+  // class variables
+  static PushingBox *_instance;
+
+  // instance variables
+  EthernetClient client;
+  const char *DevID = "v172D35C152EDA6C";  // DeviceID from Pushingbox
+  unsigned long nextSendTime = 0;
+  byte server[4] = {64, 233, 187, 99};
```
jgfoster, line 28: In production the server should be `api.pushingbox.com`. We should think about how the mock server will handle that...

jgfoster (changes requested):

jgfoster (approved):

prestoncarman (approved):
LGTM

PR #172, opened 2021-05-28
Update returnCurrentValue() to return current value

eucalvo:
return current value for most UI setters.

### DIFF
--- a/test/SetGoogleSheetIntervalTest.cpp
+++ b/test/SetGoogleSheetIntervalTest.cpp
@@ -12,7 +12,7 @@ unittest(test) {
   tc->setNextState(test, true);
 
   // getCurrentValue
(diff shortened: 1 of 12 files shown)


test/SetGoogleSheetIntervalTest.cpp
```diff
@@ -11,6 +11,9 @@ unittest(test) {
   SetGoogleSheetInterval* test = new SetGoogleSheetInterval(tc);
   tc->setNextState(test, true);
 
+  // getCurrentValue
+  assertEqual(65535, test->getCurrentValue());
```
prestoncarman, line 15: Might be nice to add a test for when current value actually has been set.

jgfoster: Done.

jgfoster (approved):

PR #180, opened 2021-06-03
Fix #117

### LINKED ISSUE #117: Add PID enabled to ViewPID display
We are able to enable/disable the PID, and the flag is used when computing the on-time for the CO2 bubbler, but we don't have a UI to view the state. We do have three lines that show the Kp, Ki, and Kd values. This would be a good place to add that view.

Paydenb:
Fix problem#117:
Now there is a UI that displays if PID is on or off

### DIFF
--- a/src/UIState/SeePIDConstants.h
+++ b/src/UIState/SeePIDConstants.h
@@ -24,7 +24,6 @@ private:
   void loadKi(uint16_t line);
   void loadKd(uint16_t line);
   void loadPID(uint16_t line);
-//   void loadSlope(uint16_t line);
 
   uint32_t startTime;
 };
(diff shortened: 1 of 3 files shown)


src/UIState/SeePIDConstants.h
```diff
@@ -23,7 +23,8 @@ class SeePIDConstants : public UIState {
   void loadKp(uint16_t line);
   void loadKi(uint16_t line);
   void loadKd(uint16_t line);
-  void loadSlope(uint16_t line);
+  void loadPID(uint16_t line);
+//   void loadSlope(uint16_t line);
```
jgfoster, line 27: If we find we don't need this file (things work with it commented out), let's remove the line.

jgfoster (changes requested):
Delete one line...

src/UIState/SeePIDConstants.h
jgfoster: Note also that this line fails the formatting test, so when it is removed we should have passing tests!

jgfoster (approved):

prestoncarman (approved):
LGTM

prestoncarman:
@all-contributors please add @Paydenb for code

PR #181, opened 2021-06-04
Don't trim last character read from stream unless it is a '\r'

### LINKED ISSUE #178: pH data appears to be rounded to hundredths (missing thousandths value)
While we report pH with three digits after the decimal, the third (thousandths) is always zero, which seems exceedingly unlikely!

jgfoster:
Fix #178.


prestoncarman (approved):
LGTM

Issue #179, opened 2021-06-03
PushingBox pH value is often zero


jgfoster:
Some devices show occasional/frequent/continuous values of 0 for pH in what is sent to PushingBox. This appears to be related to the fact that the first transmission to PushingBox is made before we have a pH reading. While we should figure out why the reset is happening so often, the current work-around is to wait more than a minute before sending the first report.

jgfoster:
We believe that this was fixed by #183.

PR #183, opened 2021-06-04
Wait 70 seconds before sending first PushingBox report

jgfoster:
Possible fix for #179

### DIFF
--- a/src/Devices/PushingBox.cpp
+++ b/src/Devices/PushingBox.cpp
@@ -53,12 +53,6 @@ void PushingBox::sendData() {
     serial("Set Tank ID in order to send data to PushingBox");
     return;
   }
-  serial("attempting to connect to PushingBox...");
-  if (!client.connected() && !client.connect(server, 80)) {
-    serial("connection failed");
-    return;
-  }
-  serial("connected");
   char format[] =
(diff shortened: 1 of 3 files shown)


src/Devices/PushingBox.cpp
```diff
@@ -69,5 +63,12 @@ void PushingBox::sendData() {
   float temperature = TempProbe_TC::instance()->getRunningAverage();
   float pH = PHProbe::instance()->getPh();
   snprintf(buffer, sizeof(buffer), format, DevID, tankID, temperature, pH);
-  client.write(buffer, strnlen(buffer, sizeof(buffer)));
+  serial(buffer);
+  serial("attempting to connect to PushingBox...");
+  if (client.connected() || client.connect(server, 80)) {
```
prestoncarman, line 68: Why move this code to later in the function?

prestoncarman (approved):
LGTM

src/Devices/PushingBox.cpp
jgfoster: I wanted to see the data that was generated to send to PushingBox even if we could not connect. The problem being investigated was that we often have zero and this allowed me to see what would have been sent.

PR #188, opened 2021-06-05
Collection of small changes plus version

### LINKED ISSUE #170: Startup option to bypass network
At startup show a countdown timer for 10 seconds and offer the following options:

A) Attempt to connect to Ethernet and get a DHCP-assigned IP address (will hang for up to 60 seconds)
D) Disregard the DHCP and use a self-assigned IP address

The self-assigned IP address should be 192.168.1.X where X is the Tank ID.

### LINKED ISSUE #185: Display IP address is always 000.000.000.000
Something is different from the tests!

jgfoster:
- Edit menu to use "View Settings" rather than "View TC Settings"
- Update output format for writing to the SD card
- Update test for temperature and pH decimal precision
- Save the DHCP-assigned IP address for later display (fix #185)
- Shorten the Ethernet startup timeout to 5 seconds (fix #170)
- Update version to 21-05-4

### DIFF
--- a/test/TCLibTest.cpp
+++ b/test/TCLibTest.cpp
@@ -1,7 +1,9 @@
 #include <Arduino.h>
 #include <ArduinoUnitTests.h>
 
+#define DEBUG 1
 #include "DateTime_TC.h"
+#include "EEPROM_TC.h"
 #include "PHControl.h"
 #include "PHProbe.h"
 #include "SD_TC.h"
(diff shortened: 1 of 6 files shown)


test/TCLibTest.cpp
```diff
@@ -109,13 +109,13 @@ unittest(storeDataToSD) {
   contents = contents.substring(i + 1);
   i = contents.indexOf('\n');
   line = contents.substring(0, i);
-  String expected("04/27/2021 14:24:50,   0, 20.021, 20.000, 7.5000, 7.5000,    0, 100000.0,   0.0,   0.0");
+  String expected("04/27/2021 14:24:50,   0, 20.02, 20.00, 7.500, 7.500,    0, 100000.0,      0.0,      0.0");
```
prestoncarman, line 112: When thinking about printing significant digits and our recent issue #178, I think we should consider using test values that have unique values for all decimal places. Zero may be from padding or having a value that is zero.

prestoncarman (changes requested):
Can we update the test to include values besides zero in the decimal values.

test/TCLibTest.cpp
jgfoster: I believe I've addressed this. Could you review it and let me know?

prestoncarman (approved):
LGTM

test/TCLibTest.cpp
prestoncarman: Nice job.

PR #196, opened 2021-06-09
Consolidate IP and Mac address view (fix #191)

### LINKED ISSUE #191: Consolidate IP and Mac Address View
We have two menu commands that take us to the same view. One menu command should be enough!

jgfoster:


prestoncarman (approved):
LGTM

PR #205, opened 2021-06-10
PushingBox device ID now in TankController.ino

jgfoster:
Provide PushingBox device ID in top-level sketch instead of lower. This will need to be edited before compile and install.


test/PushingBoxTest.cpp
```diff
@@ -60,21 +72,19 @@ unittest(SendData) {
...
       "\r\n";
-
   assertEqual(expected1, bufferResult.c_str());
   char expected2[] =
-      "GET /pushingbox?devid=v172D35C152EDA6C&tankid=99&tempData=20.26&pHdata=7.125 HTTP/1.1\r\n"
-      "Host: api.pushingbox.com\r\n"
-      "Connection: close\r\n"
-      "\r\n"
-      "\r\n"
+      // "15:26 pH=7.125 temp=20.26\r\n"
```
prestoncarman, line 81: Does this need to be commented out?

prestoncarman (approved):
LGTM

PR #207, opened 2021-06-17
Use jgfoster/arduino_ci for testing (massive speed gains)

jgfoster:
I've created a branch of `arduino_ci` that includes my speedup of the testing process. This should reduce the test cycle from 42:30 to 2:30 (on a recent run).


prestoncarman (approved):
Nice speed improvement.

PR #209, opened 2021-06-30
Use seeded random number to generate MAC address

### LINKED ISSUE #192: Initialize Mac Address Properly
Ethernet requires a six-byte MAC (Media Access Control) address and we currently define it as `{ 0x90, 0xA2, 0xDA, 0x00, 0x00, 0x00 }` in `Ethernet_TC.h:33`. Branch 2019 defines the same default but during setup looks in EEPROM for a stored value. If it is not defined it generates one using the TrueRandom library. We should do the same.

EEPROM stores the MAC address in 44-51 (8 bytes). Location 44 is a flag for valid value (if the character is '#') and locations 47 to 49 represent bytes 3-5 of the MAC. Locations 45, 46, 50, and 51 are unused.

jgfoster:
Fix #192.


prestoncarman (approved):
LGTM

prestoncarman (approved):
Nice addition of the reset when viewing the MAC.

PR #212, opened 2021-07-01
Use Keypad `main` rather than `master`

jgfoster:
This will give us a better-maintained library with a LGPL 2.1 license.


prestoncarman (approved):
LGTM

PR #217, opened 2021-07-09
v21.07.2

### LINKED ISSUE #186: Production devices seem to reset frequently
In the lab the uptimes for three devices are 90 seconds, 28 minutes, and 4 hours. The development Arduino boards (with no external probes or A/C devices) will run for days without resetting.

Do the lab devices running the old code reset frequently? Is this a software problem or a hardware problem? Could we add uptime to the PushingBox reports?

jgfoster:
- Memory cleanup to avoid stack/heap collision (fix #186 ?)
- Remove many uses of String
- Add View Free Memory menu option

### DIFF
--- a/.github/actions/spelling/allow/TankControllerLib.cpp.txt
+++ b/.github/actions/spelling/allow/TankControllerLib.cpp.txt
@@ -1,0 +1,2 @@
+brkval
+bss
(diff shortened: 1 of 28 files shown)


prestoncarman (approved):
LGTM

PR #221, opened 2021-07-15
Idle screen shows 'B' when bubbling

jgfoster:
* Use TURN_SOLENOID_OFF as a constant instead of HIGH
* Use `tc` instead of `pTC` for TankControllerLib pointer (consistency)
* Rename `PIN` to `PH_CONTROL_PIN` or `TEMP_CONTROL_PIN`
* Show 'B' on idle screen when bubbler is on
* Don't add date/time to serial when changing pH Control state
* Simplify pH Control logic
* Cap bubbler on time to 9/10 seconds
* Add serial output to console during debugging
* Fix compiler warning on cast of freeMemory() value

While we haven't seen a reason the bubbler would stay on, this lets us know if the software thinks it is trying to turn it on.

### DIFF
--- a/.github/actions/spelling/allow/TankControllerLib.cpp.txt
+++ b/.github/actions/spelling/allow/TankControllerLib.cpp.txt
@@ -1,0 +1,2 @@
+brkval
+bss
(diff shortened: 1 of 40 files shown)


prestoncarman (approved):
LGTM

PR #226, opened 2021-07-15
Show temperature control in idle screen

jgfoster:
Lowercase 'h' or 'c' means turned off while uppercase 'H' or 'C' means turned on.


prestoncarman (approved):
LGTM

PR #228, opened 2021-07-20
Wait 60 seconds for DHCP

jgfoster:
Revert to original code since the DHCP request seems to fail frequently.


prestoncarman (approved):
LGTM

PR #229, opened 2021-07-20
Trigger a reset if the pH goes below 7.0

jgfoster:
So bubbler isn't on continually.

### DIFF
--- a/src/Devices/PHProbe.cpp
+++ b/src/Devices/PHProbe.cpp
@@ -71,7 +71,8 @@ void PHProbe::serialEvent1() {
           wdt_disable();
           wdt_enable(WDTO_15MS);
           serial("pH value dropped to %5.3f so trigger a reset!", value);
-          while (true) {}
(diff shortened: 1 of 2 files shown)


src/Devices/PHProbe.cpp
```diff
@@ -64,6 +66,14 @@ void PHProbe::serialEvent1() {
       if (isdigit(string[0])) {  // if the first character in the string is a digit
         // convert the string to a floating point number so it can be evaluated by the Arduino
         value = string.toFloat();
+        // we have seen situations where the CO2 bubbler stays on and drives the pH down
+        if (value < 7.0) {  // hang so as to trigger the watchdog timer reset
```
prestoncarman, line 70: While this could be hard coded, this seems like a threshold that should be user defined. Maybe we open an issue for customization.

prestoncarman (approved):
LGTM

PR #239, opened 2021-08-17
Define `SS` for `SdFat` library

jgfoster:
This was missed in the transition from the 'SD' library to the 'SdFat' library.

### DIFF
--- a/src/Devices/SD_TC.h
+++ b/src/Devices/SD_TC.h
@@ -1,7 +1,7 @@
 #pragma once
 
 #include <Arduino.h>
-#include <SD.h>
+#include <SdFat.h>
 
 typedef void (*visitor)(File* pEntry, String parentPath);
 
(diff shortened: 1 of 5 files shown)


src/Devices/SD_TC.h
```diff
@@ -1,7 +1,9 @@
 #pragma once
 
 #include <Arduino.h>
-#include <SD.h>
+
+#define SS 4
```
prestoncarman, line 5: Why do we need the define statement?

jgfoster: Somehow it wasn't needed for (or was defined by) the SD library (which is GPL). When we switched to using the SdFat library (LGPL) there was a compile error that `SS` was not defined. I suspect that it is something that is defined by one of the real Arduino header files but is not defined by the Arduino_CI header files. When looking at its sole use, it is the default for a constructor (or begin) that specifies a pin. I looked up what we were using and assigned it. That was as far as I got.

prestoncarman (approved):
LGTM

PR #242, opened 2021-08-25
Fix compiler warning when doing free space calculations

jgfoster:
This was noticed when uploading the application to the device.


prestoncarman (approved):
LGTM

PR #243, opened 2021-08-25
Use Mozilla License

jgfoster:


prestoncarman:
Why use a PDF instead of plan text?

jgfoster:
The ["official" version](https://www.mozilla.org/en-US/MPL/2.0/) has nice formatting including highlight color for sections 6 and 7 and I thought it was appropriate to preserve that as much possible. We could provide a web page with formatting, but this seemed consistent with what they had. But I don't care. Can you find a file we should use instead?

prestoncarman:
PR  #246 shows the txt license file created using the GitHub license picker.

PR #245, opened 2021-08-25
Update for changes to Ethernet library

jgfoster:
* Skip `Ethernet.maintain()` if we did not get an IP address.
* Changes to PushingBox based on changes to Ethernet library.


prestoncarman (approved):
Did the Ethernet library get updated?

jgfoster:
See [here](https://github.com/Arduino-CI/Ethernet/commits/master) for the most recent commits to the Ethernet library used for test and production.

Issue #241, opened 2021-08-24
Use Mozilla Public License, version 2.0


jgfoster:
https://www.mozilla.org/en-US/MPL/2.0/

PR #246, opened 2021-08-25
Create LICENSE

prestoncarman:
This is the MPL v2 created using the Github license picker.


jgfoster (approved):
LGTM. Once we have this merged I can update my PR to remove the PDF and then we can get the badges and other things that were in #241.

PR #253, opened 2021-08-27
Check for a drop in free space

jgfoster:
In essence, it will force us to review the change and adjust this test if we want to allow the memory use.


prestoncarman (approved):
LGTM

PR #256, opened 2021-09-05
Stop using %f for printing floats

### LINKED ISSUE #247: Replace printf() use of float variables
This requires a separate library and the instructions for setup are a bit of a burden. It would be nice if we could just convert floats to an integer and fractional portion and print them separately (see [here](https://forum.arduino.cc/t/how-do-you-convert-a-float-to-string-solved/237090/6)).

jgfoster:
Fix #247.


prestoncarman (approved):
LGTM

Issue #255, opened 2021-08-31
Add TankController library to Arduino repository


jgfoster:
It would be very nice if new users could install the TankController software from the Arduino IDE without having to go to our GitHub releases. See [here](https://github.com/arduino/library-registry/).

jgfoster:
See https://github.com/arduino/library-registry/pull/477.

jgfoster:
I think this is done!

PR #257, opened 2021-09-05
Follow Structure for Arduino Library

jgfoster:
* Use TankController instead of TankControllerLib so that name fits within suggested size limit.
* Update `library.properties` for submission to Arduino library list.


jgfoster:
Needed for #255.

jgfoster:
Partial fix for #255.

prestoncarman (approved):
Nice name clean up.

PR #260, opened 2021-09-05
Include size report with each PR

### LINKED ISSUE #139: Display program size and memory usage on PR
When the program is compiled with Arduino CLI, the compiler shares a few stats that would be nice to know about the change being requested. Specifically:

* Program size (X out of 256K)
* Memory size (X out of 8K)

jgfoster:
The arduino_ci test updates `size.txt` when it builds the sketch. Thus each commit will include the current values (if the tests are run!). You can see the new size by viewing the modified files. To ensure that the tests are run before a PR, part of the GitHub action is to update the file and see if it changes. See [here](https://github.com/jgfoster/TankController/runs/3518130868?check_suite_focus=true) for a sample failure.

Fix #139.

### DIFF
--- a/.github/workflows/Arduino-CI.yaml
+++ b/.github/workflows/Arduino-CI.yaml
@@ -19,3 +19,5 @@ jobs:
           g++ -v
           scripts/install_libraries.sh
           scripts/testAndBuild.sh
+      - name: Ensure no Git changes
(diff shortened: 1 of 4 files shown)


prestoncarman (approved):
LGTM

PR #261, opened 2021-09-07
Remove unused file/code

jgfoster:
Since we refresh the LCD every loop, there isn't any need for a custom menu to do this.


prestoncarman (approved):
LGTM

PR #262, opened 2021-09-15
Change Arduino Library Lint to recognize we are in the public library listing

jgfoster:
This should avoid the error we get now that the name is a duplicate. See #255.


prestoncarman (approved):
LGTM

PR #266, opened 2021-11-04
Add ramp function for pH

eucalvo:


src/Devices/PHControl.cpp
```diff
@@ -66,9 +92,16 @@ bool PHControl::isOn() {
 void PHControl::updateControl(float pH) {
   int msToBeOn;
   int nowModWindow = millis() % WINDOW_SIZE;
+  float currentTime = DateTime_TC::now().secondstime();
+  currentPHTarget =
+      rampStartingPh + ((currentTime - rampTimeStart) * (targetPh - rampStartingPh) / (rampTimeEnd - rampTimeStart));
+  // if ramp is not used
+  if (currentTime >= rampTimeEnd) {
```
jgfoster, line 99: I suggest and if-else with the condition being `currentTime < rampTimeEnd`.

src/Devices/PHControl.cpp
```diff
@@ -34,6 +35,13 @@ PHControl::PHControl() {
     targetPh = DEFAULT_PH;
     EEPROM_TC::instance()->setPH(targetPh);
   }
+  rampTimeEnd = EEPROM_TC::instance()->getRampTimeEnd();
+  if (rampTimeEnd == 0xFFFFFFFF || rampTimeEnd == 0) {
```
jgfoster, line 39: If the end is set to a valid value, then we should load the other values. This is for the use case where the device is reset during the ramp. One also might consider clearing the values if we are beyond the end time.

test/PHControlTest.cpp
```diff
@@ -169,4 +169,27 @@ unittest(disableDuringCalibration) {
...
+  delay(1800000);  // delay 30 minutes
+  tc->loop();
+  assertTrue(8.0 <= controlSolenoid->getCurrentPHTarget() && controlSolenoid->getCurrentPHTarget() <= 8.01);
+  delay(1800000);  // delay 30 minutes
+  tc->loop();
+  assertTrue(7.5 <= controlSolenoid->getCurrentPHTarget() && controlSolenoid->getCurrentPHTarget() <= 7.51);
+  delay(1800000);  // delay 30 minutes
+  tc->loop();
+  assertEqual(7, controlSolenoid->getCurrentPHTarget());
+  // ramp time no longer used after it ends
```
jgfoster, line 188: Nice tests!

jgfoster (changes requested):
See notes...

jgfoster (approved):

PR #269, opened 2021-12-03
Sine wave pH

eucalvo:


jgfoster (approved):

src/Devices/EEPROM_TC.cpp
```diff
@@ -126,6 +135,9 @@ uint32_t EEPROM_TC::getRampTimeEndTemp() {
 float EEPROM_TC::getRampStartingTemp() {
   return eepromReadFloat(RAMP_STARTING_TEMP_ADDRESS);
 }
+uint16_t EEPROM_TC::getPHSetType() {
```
prestoncarman, line 138: "Ph" or "PH"? Lets be consistent and make it always "Ph" in the function names.

src/Devices/EEPROM_TC.cpp
```diff
@@ -213,6 +234,9 @@ void EEPROM_TC::setRampTimeEndTemp(uint32_t value) {
 void EEPROM_TC::setRampStartingTemp(float value) {
   eepromWriteFloat(RAMP_STARTING_TEMP_ADDRESS, value);
 }
+void EEPROM_TC::setPHSetType(uint16_t value) {
```
prestoncarman, line 237: "Ph" or "PH"? Lets be consistent and make it always "Ph" in the function names.

src/Devices/EEPROM_TC.h
```diff
@@ -26,16 +26,20 @@ class EEPROM_TC {
...
+  uint32_t getPhSinePeriod();
+  float getPhSineAmplitude();
   uint32_t getRampTimeStartTemp();
   uint32_t getRampTimeEndTemp();
   float getRampStartingTemp();
-  float getPHDelay();            // not used
-  float getPHInterval();         // not used
-  float getPHSeriesPointer();    // not used
-  float getPHSeriesSize();       // not used
+  float getPHDelay();          // not used
```
prestoncarman, line 38: "Ph" or "PH"? Lets be consistent and make it always "Ph" in the function names. Update for the whole file.

test/MenuTest.cpp
```diff
@@ -37,6 +37,11 @@ unittest_teardown() {
   enterKey('D');
 }
 
+// 2 goes up in the menu
```
prestoncarman, line 40: I think this comment would be nice to have as a multiline comment associated with the function.

```c++
/**
 * 2 goes up in the menu
 * ...
 */
unittest(MainMenu) {
```

This should help it to show up with function documentation.

prestoncarman (changes requested):
The main question is how to deal with naming pH variables and files. I think we need to be more consistent. We seem to use "pH", "PH", and "Ph". I can see arguments for all of these, but in most cases I think that "Ph" is probably the correct answer.
Thought?

jgfoster:
The chemists and biologists use "pH" as the proper designation, but that doesn't fit very well with camelCase. So, consistency is good and Ph fits so I'd be inclined to go with that...

prestoncarman (approved):

jgfoster (approved):

PR #271, opened 2022-01-12
Fix test failures on code.cs

pianistrevor:
Fixes an error in free memory when running test suite on different location (GitHub vs code.cs)
Fixes an error where the rtc on Tank Controller was not being updated before the state was changed.


jgfoster (approved):

prestoncarman (approved):
LGTM

Issue #274, opened 2022-02-04
Show IP address to serial port on startup


jgfoster:
It tells us that it is listening on port 80 but it would be nice to know how to reach it!

pianistrevor:
Fixed in 4fc9b01769082887a96d4f6a1aed865e4b74b347

PR #276, opened 2022-02-04
Fix for #274

pianistrevor:
Add IP Address to serial message in EthernetServerTC constructor

### DIFF
--- a/src/Devices/EthernetServer_TC.cpp
+++ b/src/Devices/EthernetServer_TC.cpp
@@ -3,6 +3,9 @@
 #include <avr/wdt.h>
 
 #include "DateTime_TC.h"
+#include "Devices/Ethernet_TC.h"
+#include "Devices/JSONBuilder.h"
+#include "Devices/LiquidCrystal_TC.h"
 #include "SD_TC.h"
 #include "Serial_TC.h"
 #include "TankController.h"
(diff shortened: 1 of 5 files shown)


src/Devices/EthernetServer_TC.cpp
```diff
@@ -27,7 +29,8 @@ EthernetServer_TC* EthernetServer_TC::instance() {
  */
 EthernetServer_TC::EthernetServer_TC(uint16_t port) : EthernetServer(port) {
   begin();
-  serial(F("Ethernet Server is listening on port 80"));
+  IP Address = Ethernet_TC::instance()->getIP();
+  serial(F("Ethernet Server is listening on %i.%i.%i.%i:%i"), IP[0], IP[1], IP[2], IP[3], getPort());
```
jgfoster, line 33: Codacy points out that the variable Address is never used. I think `IP` is the "type" and `Address` is the "name".

src/Devices/EthernetServer_TC.cpp
```diff
@@ -3,6 +3,8 @@
 #include <avr/wdt.h>
 
 #include "DateTime_TC.h"
+#include "Devices/Ethernet_TC.h"
+#include "Devices/LiquidCrystal_TC.h"
```
jgfoster, line 7: Why was LiquidCrystal_TC.h added?

jgfoster (commented):
Correction requested.

src/Devices/EthernetServer_TC.cpp
pianistrevor: I closed this yesterday due to these problems. Check the open one.

PR #277, opened 2022-02-05
Fix for #274

pianistrevor:
Add IP Address to serial message in EthernetServer_TC constructor

### DIFF
--- a/size.txt
+++ b/size.txt
@@ -1,2 +1,1 @@
-Sketch uses 77988 bytes (30%) of program storage space. Maximum is 253952 bytes.
 Global variables use 2224 bytes (27%) of dynamic memory, leaving 5968 bytes for local variables. Maximum is 8192 bytes.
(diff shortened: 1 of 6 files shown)


size.txt
```diff
@@ -1,2 +1,2 @@
-Sketch uses 77946 bytes (30%) of program storage space. Maximum is 253952 bytes.
+Sketch uses 77988 bytes (30%) of program storage space. Maximum is 253952 bytes.
```
jgfoster, line 1: We don't really care about the change in the size of the program, only about the change in global variables. So including this line is unnecessary and adds confusion. Please modify `scripts/testAndBuild.sh` so that line 6 reads as follows:

```
tail -n 3 output.txt | head -n 1 > size.txt
```

Then submit that file and a new (modified) `size.txt`. Otherwise, this looks good.

jgfoster (changes requested):

size.txt
pianistrevor: Changes implemented.

jgfoster: Hmm... I expected a new size file with only one line...

pianistrevor: So did I. For some reason the new size.txt wasn't added to my commit. Doing that now.

jgfoster (approved):

PR #280, opened 2022-02-15
Release 22.02.1

jgfoster:
* Update version to 22.02.1 
* Edit paths so VS Code can find include files 
* Add a couple defines so VS Code is happy


prestoncarman (approved):
LGTM

PR #283, opened 2022-02-15
Fix buffer overrun problem

jgfoster:
A pH of 10.000 or more does not print nicely in a space formatted for 5.3.


prestoncarman (approved):
LGTM

PR #284, opened 2022-02-16
EthernetServer API

pianistrevor:
- Handle GET and POST requests for the display and keypresses
- Return a JSON file with all current values
- Add test coverage for each


jgfoster:
It looks like the JSON Builder is expecting the old version. We should look at a way to simplify updating the version without changing it in many places!

pianistrevor:
Yes, I hard coded it into the test. We can either change it there or do more complex string interpolation with the static version number

test/EthernetServerTest.cpp
```diff
@@ -44,8 +46,123 @@ unittest(basic) {
...
+    response.concat(pBuffer->front());
+    pBuffer->pop_front();
+  }
+  const char expectedResponse[] = "HTTP/1.1 400 Bad Request\r\n\r\n";
+  assertEqual(expectedResponse, response);
+  assertEqual(NOT_CONNECTED, server->getState());
+  client.stop();
+  server->loop();
+}
+
```
prestoncarman, line 167: Add test for current endpoint.

src/UIState/SeeDeviceAddress.cpp
```diff
@@ -28,15 +28,33 @@ void SeeDeviceAddress::handleKey(char key) {
   };
 }
 
+int SeeDeviceAddress::writeToBuffer(char* buffer, int size, int line) {
```
prestoncarman, line 31: * Refactor to use a ENUM to identify the type of display instead of the size of the buffer.
* Refactor to identify the information displayed to use more appropriate term instead of "line".

prestoncarman (changes requested):
Note the I only commented on one file, but it affects most of them.

test/EthernetServerTest.cpp
pianistrevor: Were you referring to the unit test for display, or did you have something else in mind?

pianistrevor: Test for current values added to EthernetServerTest.cpp

prestoncarman (approved):

PR #286, opened 2022-02-22
View ph

eucalvo:


src/UIState/SeePh.cpp
```diff
@@ -0,0 +1,105 @@
...
+  LiquidCrystal_TC::instance()->writeLine(buffer, line);
+}
+
+void SeePh::loadCurrent(uint16_t line) {
+  char buffer[17];
+  float overallTargetPh = PHControl::instance()->getTargetPh();
+  float currentTargetPh = PHControl::instance()->getCurrentPhTarget();
+  float currentPh = (int)(PHProbe::instance()->getPh() * 100 + .5);
+  currentPh = currentPh / 100;
+  snprintf_P(buffer, sizeof(buffer), (PGM_P)F("%i.%i %i.%i %i.%i"), (int)currentPh, (int)(currentPh * 100) % 100,
```
jgfoster, line 101: What if `currentPh` is `8.07`?
We get `8.` followed by (8.07 * 100) % 100, or 807 % 100, or 7, which will print as `7` rather than `07`. What if you had a format of `%02i`, or "two digits, right justified, zero-filled".
Same issue below (and wherever you found the code you copied!). See #287 for the more wide-spread problem.

test/SeePhTest.cpp
```diff
@@ -0,0 +1,162 @@
...
+PHControl *controlSolenoid = PHControl::instance();
+LiquidCrystal_TC *lc = LiquidCrystal_TC::instance();
+
+unittest_setup() {
+  DateTime_TC january(2021, 1, 15, 1, 48, 24);
+  january.setAsCurrent();
+}
+
+unittest_teardown() {
+  DateTime_TC january(2021, 1, 15, 1, 48, 24);
```
jgfoster, line 24: I don't think you need to set the date/time at the end of each test.

test/SeePhTest.cpp
```diff
@@ -0,0 +1,162 @@
...
+  char buffer[10];
+  snprintf_P(buffer, sizeof(buffer), (PGM_P)F("%i.%i\r"), (int)value, (int)(value * 1000 + 0.5) % 1000);
+  state->serialPort[1].dataIn = buffer;  // the queue of data waiting to be read
+  tc->serialEvent1();                    // fake interrupt to update the current pH reading
+  tc->loop();                            // update the controls based on the current readings
+}
+
+unittest(TestVerticalScrollWithFlatSet) {
+  setPhMeasurementTo(7.00);
+  controlSolenoid->setTargetPh(7.00);
```
jgfoster, line 38: Perhaps try some values that demonstrate the problem for #287.

jgfoster (changes requested):
See comments; also add test for ramp finished.

output.txt
```diff
@@ -0,0 +1,247 @@
+                           .                  __  ___
```
jgfoster, line 1: This file should be listed in `.gitignore` so that it is not included in the commit.

src/UIState/SeePh.cpp
```diff
@@ -0,0 +1,98 @@
...
+#include "SeePh.h"
+
+#include "Devices/DateTime_TC.h"
+#include "Devices/LiquidCrystal_TC.h"
+#include "Devices/PHControl.h"
+#include "Devices/PHProbe.h"
+#include "TC_util.h"
+
+void SeePh::loop() {
+  int32_t elapsedSeconds = ((millis() - startTime) / 3000) % 2;
```
jgfoster, line 14: Perhaps a name like `oddEven` would be better since this value is not anything close to "elapsedSeconds".

src/UIState/SeePh.cpp
```diff
@@ -0,0 +1,98 @@
...
+    }
+    case RAMP_TYPE: {
+      type = (char *)"ramp";
+      break;
+    }
+    case SINE_TYPE: {
+      type = (char *)"sine";
+      break;
+    }
+    default:
```
jgfoster, line 43: While it is unlikely that the default case would ever happen, as long as you have it you could assign a value of "????" to `type` (as it is you have a random value in that variable).

src/UIState/SeePh.cpp
```diff
@@ -0,0 +1,105 @@
...
+                   (int)timeLeftSeconds);
+        LiquidCrystal_TC::instance()->writeLine(buffer, line);
+      } else {
+        snprintf_P(buffer, sizeof(buffer), (PGM_P)F("left: %i:%i:%i"), 0, 0, 0);
+        LiquidCrystal_TC::instance()->writeLine(buffer, line);
+      }
+      break;
+    }
+    case SINE_TYPE: {
+      float period = PHControl::instance()->getPeriod();
```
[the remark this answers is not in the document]
jgfoster: It appears that the reason for making this a `float` is so that the next line will evaluate to a `float`. Another way to do this and be more explicit is to have the divisor be a `float`: `3600.0`.

test/SeePhTest.cpp
jgfoster: Please change one or both of the values to a fractional power of two.

jgfoster (changes requested):
See notes.

jgfoster (approved):
I'll ask @prestoncarman for his review as well...

jgfoster (approved):

PR #288, opened 2022-02-25
Fix multiple NumCollectorStateBugs

### LINKED ISSUE #273: Question  mark on display when entering floating-point numbers
The key sequence "A8.1" does not show correctly. As soon as you press the '1' key the number is replaced with '?'. This suggests that we are trying to use the floating-point printf() format (see #137).

pianistrevor:
- Use dtostrf() for printing entry with appropriate precision, rather than the original method of formatting a format string
- Fix bug where clearing value did not reset the factor of precision
- Add test coverage for clear to catch this

This fixes #273.


pianistrevor:
The extra commits were the result of incorrect upstream merges; these will be squashed here and I've corrected the problem in main on my fork.

jgfoster (approved):
I've asked @prestoncarman to approve as well.

prestoncarman (approved):
LGTM

PR #290, opened 2022-02-25
Remove pH failsafe

### LINKED ISSUE #231: Customize (or eliminate) pH failsafe
PR #229 provides a failsafe to reboot the device if the pH falls below 7.0. This was added in an attempt to prevent the behavior observed in 21.07.2 which left the bubbler on continually (very bad!). If we choose to leave this in the value should probably be directly configurable or a configurable value below the target pH.

pianistrevor:
Fixes #231 per request of @jgfoster


pianistrevor:
The extra commits were the result of incorrect upstream merges; these will be squashed here and I've corrected the problem in main on my fork.

jgfoster (approved):
I've asked @prestoncarman to approve as well.

Issue #224, opened 2021-07-15
Method to clear temperature calibration 


KirtOnthank:
It would be nice to have a method to clear the temperature calibration like we can clear pH calibration.

pianistrevor:
Fixed via #296

PR #296, opened 2022-03-04
Add menu for clearing temperature calibration

pianistrevor:
Implements #224


src/Devices/TempProbe_TC.cpp
```diff
@@ -89,3 +89,17 @@ void TempProbe_TC::setCorrection(float value) {
...
+ * clearCorrection(f)
+ *
+ * Sets the correction offset back to 0
+ */
+void TempProbe_TC::clearCorrection() {
+  COUT("old = " << correction);
+  if (correction != 0) {
+    correction = 0.0;
+    EEPROM_TC::instance()->setCorrectedTemp(correction);
+    serial(F("Set temperature correction to %i.%02i"), (int)correction, (int)(correction * 100 + 0.5) % 100);
```
jgfoster, line 103: You could just say that the new correction is "0.0", but this is fine as well.

jgfoster (approved):
LGTM. Let's ask @prestoncarman for a review as well...

src/Devices/TempProbe_TC.cpp
pianistrevor: I like this because we would catch it if it’s bugged.

prestoncarman (approved):
LGTM

PR #299, opened 2022-03-10
Fix WDT hanging too long on setup

pianistrevor:
Starts the WDT after printing serial messages in case additional messages that take longer than the WDT failsafe are inserted later.
Removes the printout of the root directory of the SD card.


jgfoster (approved):
Well, that wasn't too complicated!

pianistrevor:
So somehow removing a function call and reordering it adds 2 bytes of static variables? Odd

prestoncarman (approved):
LGTM, Probably should have been this way to start with.

jgfoster (approved):

PR #302, opened 2022-03-14
Fix numerous test errors

### LINKED ISSUE #301: Using 'strnlen' despite not being a standard C++ function
`TankController.cpp` uses `strnlen()` calls. This function is not defined in the Standard C++ library.

On Linux and MacOS it is included as a GNU extension in `string.h`, but we might consider defining the function ourselves. On Windows, the GNU extensions are not automatically installed, and using Cygwin will limit the platforms that `strnlen` can be executed on (specifically, POSIX). See [this](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/string.h.html) for more info.

If we want this to be limited to Linux and Mac in the future, that's fine, but I like the idea of it working across multiple environments if/when we go public.

pianistrevor:
- Fix DateTimeTest delay period for CPU-intensive environment
- Change free memory test, update tests that use this function
- Modify PHControl and TemperatureControl headers to prevent garbage data reporting
- Update set point tests to check for correct substate type
- Add support for Windows + CYGWIN combination (fixes #301)


src/Devices/PHControl.cpp
```diff
@@ -67,7 +67,13 @@ PHControl::PHControl() {
       amplitude = EEPROM_TC::instance()->getPhSineAmplitude();
       sineStartTime = EEPROM_TC::instance()->getPhSineStartTime();
       break;
+    case FLAT_TYPE:
```
jgfoster, line 70: I had understood from our discussion that these values were for debugging. What does it mean for a flat type to have a start and end time?

src/TC_util.h
```diff
@@ -1,6 +1,10 @@
 #pragma once
 #include <Arduino.h>
 
+#if defined(__CYGWIN__)
```
jgfoster, line 4: While I think this is good for now, perhaps we should discuss moving it to Arduino CI. That is, the only reason we are compiling on something other than the Arduino itself is for testing. Perhaps we should think about a "compatibility layer" for Windows.

test/SetPHSetPointTest.cpp
```diff
@@ -15,9 +15,11 @@ unittest_setup() {
 unittest(test_target_of_7_125_with_ramp_of_4_125) {
   LiquidCrystal_TC* lcd = LiquidCrystal_TC::instance();
   EEPROM_TC::instance()->setPh(8.125);
+  // EEPROM_TC::instance()->setPhSetType(RAMP_TYPE);
```
jgfoster, line 18: Can this be removed?

jgfoster (changes requested):

src/TC_util.h
pianistrevor: Or I can just switch the compiler in an Arduino CI PR.

jgfoster (approved):

jgfoster (approved):
Requesting review from @prestoncarman as well...

prestoncarman (approved):
LGTM

PR #310, opened 2022-04-28
Fix MainMenu excess code; add tests for read/write speeds

pianistrevor:
MainMenu's `idle()` function was calculating values every loop instead of every second, so now `TankController::loop()` executes in 2-4 ms at MainMenu idle.

This also has comments that document experimental speeds for reading from the SD card and writing to the client, in cases for handling file transfer.

This also refactors the file transfer protocol to be non-blocking and is transmitted as multipart data rather than text.

### DIFF
--- a/src/Devices/EthernetServer_TC.h
+++ b/src/Devices/EthernetServer_TC.h
@@ -9,7 +9,7 @@
 #include <Ethernet.h>
 #endif
 
-enum serverState_t { NOT_CONNECTED, READ_REQUEST, GET_REQUEST, POST_REQUEST, IN_PROGRESS, FINISHED };
+enum serverState_t { NOT_CONNECTED, READ_REQUEST, GET_REQUEST, POST_REQUEST, IN_PROGRESS, IN_TRANSFER, FINISHED };
 
 /**
  * EthernetServer_TC provides wrapper for web server for TankController
@@ -28,7 +28,7 @@ public:
     return state;
   }
   void loop();
-  void writeBufferToClient(char*, bool);
+  void writeToClientBuffer(char*, bool);
 
 private:
   // class variables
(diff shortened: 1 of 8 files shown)


src/UIState/MainMenu.cpp
```diff
@@ -236,37 +236,40 @@ void MainMenu::selectSet() {
...
-  float temp = tempProbe->getRunningAverage();
-  if (temp < 0.0) {
-    temp = 0.0;
-  } else if (99.99 < temp) {
-    temp = 99.99;
-  }
-  char status = tempControl->isHeater() ? 'h' : 'c';
-  if (tempControl->isOn()) {
-    status = toupper(status);
+  if (millis() - lastDisplayTime > 1000) {
```
jgfoster, line 239: I had expected that instead of only updating the display every second, we would modify `LiquidCrystal_TC::writeLine()` so that if the new text was the same as the old text we would skip the write. I guess I'm curious to know if the calculation of the output is expensive or the final `print()` statement in `LiquidCrystal_TC.cpp`.

src/Devices/EthernetServer_TC.h
```diff
@@ -38,23 +38,32 @@ class EthernetServer_TC : public EthernetServer {
   EthernetClient client;
   serverState_t state = NOT_CONNECTED;
   char buffer[512];
+  char boundary[13];
```
jgfoster, line 41: Should this be a size of 9?

src/UIState/MainMenu.cpp
pianistrevor: Currently the LCD `print()` is called twice per second in MainMenu, so it doesn't seem worse than calling it twice in a second when the text changes. Regardless of whether the bottleneck is in the LCD `print()` or the calculation, it seems to follow that we wouldn't want to calculate things more than we need to; in this case I think it's appropriate to say, "If we're not going to display more than one screen per second, we might as well only calculate one per second."

jgfoster: First, the idle menu might update more often than once per second, e.g., turning the bubbler on and off (though this is a minor matter). Second, when we leave the idle menu we go back to updating the display every time through the loop, so we are back to adding unnecessary delays. 

If making this change seems out of scope, then just create an issue and reference it from this one. I'd be fine with that.

jgfoster (approved):
Reviewed with Preston

PR #314, opened 2022-06-01
Arbitrary path ph

eucalvo:


src/Devices/SD_TC.cpp
```diff
@@ -45,6 +45,22 @@ void SD_TC::appendData(const char* header, const char* line) {
...
+  File file = sd.open("arb_pH_points", O_CREAT | O_WRONLY | O_APPEND);
+  file.write((void*)&value, sizeof(value));
+  file.close();
+}
+
+float SD_TC::readPhPoint(const uint32_t seekTo) {
+  float value;
+  File file = open("arb_pH_points", O_RDONLY);
+  file.seek(seekTo);
+  file.read((void*)&value, sizeof(value));
```
jgfoster, line 58: Add `// Flawfinder: ignore` to end of this line to silence the Codacy warning.

src/Devices/EthernetServer_TC.cpp
```diff
@@ -420,3 +438,61 @@ int EthernetServer_TC::weekday(int year, int month, int day) {
...
+      return;
+    }
+
+    if (next == ',' || next == ']') {
+      buffer[bufferContentsSize - 1] = '\0';
+      SD_TC::instance()->writePhPoint(charArrToFloat(buffer + 1));
+      bufferContentsSize = 1;
+      if (next == ']') {
+        PHControl::instance()->setArbitrary();
+        static const char response[] PROGMEM =
```
jgfoster, line 459: I like how `sendBadBody()` is its own method. I think that lines 459-466 should be in a similar function `sendCreated()`.

jgfoster (changes requested):
Let's talk about the test failures and see what we can do to resolve them.

jgfoster:
We will submit another PR for this.

PR #326, opened 2022-08-12
Listing files with valid HTTP header

je-foster:
This fixes some issues with HTTP headers for file transfers and for rootdir(). For the rootdir() HTTP header, the size of the body is calculated in advance by counting the number of files in the root directory on the SD card.

### DIFF
--- a/library.properties
+++ b/library.properties
@@ -1,5 +1,5 @@
 name=TankController
-version=22.7.2
+version=22.8.1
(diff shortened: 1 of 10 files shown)


jgfoster (approved):
Eventually we should look at replacing `isDoneCountingFiles` with a new state.

prestoncarman:
@all-contributors please add @je-foster for tests and code

PR #333, opened 2022-08-24
Updates

je-foster:
* New HTTP response codes: 404 (not found), 405 (not permitted), 408 (timeout), 500 (internal server error)
* Consolidate HTTP response functions for codes other than 200.
* New connections with no request remain open for 5 seconds before timing out with 408 response. Includes a test.
* If the root directory on the SD card fails to open during rootdir(), a 500 response is sent to the client and the state is reset.
* If EthernetServer_TC's loop() encounters an unknown state (due to programmer error), a 500 response is sent to the client and the state is reset.
* New state COUNTING_FILES replaces boolean isDoneCountingFiles.

### DIFF
--- a/src/Devices/EthernetServer_TC.cpp
+++ b/src/Devices/EthernetServer_TC.cpp
@@ -457,9 +457,9 @@ void EthernetServer_TC::sendResponse(int code) {
   static const char response_501[] PROGMEM =
       "HTTP/1.1 501 Not Implemented\r\n"
       "\r\n";
-  static const char *const response_table[] PROGMEM = {response_303, response_400, response_404, response_405,
+  static const char *const response_table[] = {response_303, response_400, response_404, response_405,
                                                        response_408, response_501, response_500};
-  char buffer[sizeof(response_303)];
+  char buffer[sizeof(response_303) + 1];
   switch (code) {
     case HTTP_REDIRECT:
       strncpy_P(buffer, (char *)pgm_read_word(&(response_table[0])), sizeof(buffer));
--- a/src/Devices/EthernetServer_TC.cpp
+++ b/src/Devices/EthernetServer_TC.cpp
@@ -457,9 +457,9 @@ void EthernetServer_TC::sendResponse(int code) {
   static const char response_501[] PROGMEM =
       "HTTP/1.1 501 Not Implemented\r\n"
       "\r\n";
-  static const char *const response_table[] = {response_303, response_400, response_404, response_405,
+  static const char *const response_table[] PROGMEM = {response_303, response_400, response_404, response_405,
                                                        response_408, response_501, response_500};
-  char buffer[sizeof(response_303) + 1];
+  char buffer[sizeof(response_303)];
   switch (code) {
     case HTTP_REDIRECT:
       strncpy_P(buffer, (char *)pgm_read_word(&(response_table[0])), sizeof(buffer));
--- a/src/Devices/EthernetServer_TC.cpp
+++ b/src/Devices/EthernetServer_TC.cpp
@@ -235,6 +235,11 @@ void EthernetServer_TC::sendHeadersForRootdir(int fileCount) {
 #endif
 }
 
+void EthernetServer_TC::sdError() {
+  sendErrorHeaders();
+  state = FINISHED;
+}
+
 // Tests speed for reading a file from the SD Card
 // Empirical results show about 1 ms per 512 B
 void EthernetServer_TC::testReadSpeed() {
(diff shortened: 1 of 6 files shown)


src/Devices/EthernetServer_TC.cpp
```diff
@@ -419,29 +429,60 @@ void EthernetServer_TC::sendHeadersWithSize(uint32_t size) {
...
+      "\r\n";
+  static const char response_501[] PROGMEM =
+      "HTTP/1.1 501 Not Implemented\r\n"
+      "\r\n";
+  static const char *const response_table[] PROGMEM = {response_303, response_400, response_404, response_405,
+                                                       response_408, response_501, response_500};
+  char buffer[sizeof(response_303)];
+  switch (code) {
+    case HTTP_REDIRECT:
+      strncpy_P(buffer, (char *)pgm_read_word(&(response_table[0])), sizeof(buffer));
```
jgfoster, line 465: Try `(PGM_P)response_303` instead of `&(response_table[0])` (and equivalent adjustments below, removing the `response_table`).

jgfoster (changes requested):
Very nice refactoring and cleanup of HTTP responses. I'd like one change in how SD errors are reported back to the caller. And as to size, I don't see anything that could be moved to program memory, so just edit `size.txt` to match the new size found by GitHub.

src/Devices/EthernetServer_TC.cpp
```diff
@@ -419,29 +429,58 @@ void EthernetServer_TC::sendHeadersWithSize(uint32_t size) {
...
+  static const char response_501[] PROGMEM =
+      "HTTP/1.1 501 Not Implemented\r\n"
+      "\r\n";
+  char buffer[sizeof(response_303)];
+  switch (code) {
+    case HTTP_REDIRECT:
+      strncpy_P(buffer, (PGM_P)response_303, sizeof(buffer));
+      break;
+    case HTTP_BAD_REQUEST:
+      strncpy_P(buffer, (PGM_P)response_400, sizeof(buffer));
```
prestoncarman, line 466: Does this mean the copy will always copy the same size buffer? Basically defaulting to the largest response text. Does it matter there is blank space?

src/Devices/EthernetServer_TC.cpp
```diff
@@ -189,30 +188,31 @@ void countFilesCallback(int fileCount) {
   EthernetServer_TC::instance()->sendHeadersForRootdir(fileCount);
 }
 
-// List the root directory to the client
-void EthernetServer_TC::rootdir() {
```
prestoncarman, line 193: Splitting this function makes the code much more readable. :+1:

src/Devices/EthernetServer_TC.cpp
```diff
@@ -419,29 +429,58 @@ void EthernetServer_TC::sendHeadersWithSize(uint32_t size) {
...
   client.write('\r');
   client.write('\n');
-  state = FINISHED;  // TODO: Why?! This is awkward when we want to send a body next.
 }
 
-// 303 response
-void EthernetServer_TC::sendRedirectHeaders() {
-  static const char response[] PROGMEM =
+void EthernetServer_TC::sendResponse(int code) {
+  static const char response_303[] PROGMEM =
```
prestoncarman, line 435: Is this the appropriate place for these constants? Or would it be better in the switch?

Feel free to disagree with me.

src/Devices/EthernetServer_TC.cpp
jgfoster: The third parameter to `strncpy_P()` does not specify how much will be copied but the size of the target buffer (to avoid buffer overrun bugs). The amount copied will be up to the null terminator or the size of the buffer, whichever is less.

src/Devices/EthernetServer_TC.cpp
jgfoster: I think that keeping the actual text out of the switch statement keeps the switch statement smaller and more explicit.

src/Devices/EthernetServer_TC.cpp
je-foster: Thanks, I've made these changes and it helped with the tests.

src/Devices/SD_TC.cpp
```diff
@@ -1,6 +1,7 @@
 #include "Devices/SD_TC.h"
 
 #include "DateTime_TC.h"
+#include "EthernetServer_TC.h"
```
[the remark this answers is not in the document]
je-foster: Thank you, your suggestion is a big improvement. Also, it appears to reclaim the 6 bytes.

I think a couple of the other functions return `true` if they run successfully and `false` if there is an error. May I use this convention?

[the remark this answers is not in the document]
jgfoster: Sure, that's fine. If we need more results then we can switch to error codes.

src/Devices/EthernetServer_TC.h
```diff
@@ -49,7 +49,6 @@ class EthernetServer_TC : public EthernetServer {
   void loop();
   void writeToClientBuffer(char*, bool);
   void sendHeadersForRootdir(int);
-  void sdError();
```
jgfoster: Evidently each method takes six bytes of global memory. Good to know!

src/Devices/EthernetServer_TC.cpp
je-foster: I am ambivalent. I thought this would keep the switch more concise. But I see that when someone wants to edit the text of a header they will need to scroll back and forth to make sure they find the correct one. Or maybe they won't need to scroll because they'll just look for the header they want to edit and they won't need to change the switch at all. ¯\_(ツ)_/¯

src/Devices/SD_TC.cpp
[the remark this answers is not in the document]
je-foster: Oh, I see.

PR #336, opened 2022-08-25
More HTTP response code tests

je-foster:
Merges updates that have been made to Open-Acidification/main


je-foster:
Enough tests have been added that I'll make a pull request to `main` instead of `ethernet`.

PR #337, opened 2022-08-25
Ethernet: More HTTP response codes

je-foster:
- New HTTP response codes: 404 (not found), 405 (not permitted), 408 (timeout), 500 (internal server error)
- Consolidate HTTP response functions for codes other than 200.
- New connections with no request remain open for 5 seconds before timing out with 408 response. Includes a test.
- If the root directory on the SD card fails to open during rootdir(), a 500 response is sent to the client and the state is reset.
- If EthernetServer_TC's loop() encounters an unknown state (due to programmer error), a 500 response is sent to the client and the state is reset.
- New state COUNTING_FILES replaces boolean isDoneCountingFiles.
- New test: GET request that receives 404 Not Found response
- New test: OPTIONS request that receives 405 Method Not Allowed response
- New test: FOO request that receives 501 Not Implemented response

### DIFF
--- a/library.properties
+++ b/library.properties
@@ -1,5 +1,5 @@
 name=TankController
-version=22.7.2
+version=22.8.1
 author=Kirt Onthank <Kirt.Onthank@wallawalla.edu>, Preston Carman <prestonc@apache.org>, James Foster <github@jgfoster.net>
 maintainer=James Foster <github@jgfoster.net>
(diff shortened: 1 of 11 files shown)


jgfoster (approved):

prestoncarman (approved):
LGTM

Thanks for adding more tests.

PR #339, opened 2022-10-23
Catching string overflows

je-foster:
These changes aim to replace strcpy(), strncpy(), and dtostrf(), which can cause overflows.
- strcpy() does not respect destination size at all, and easily overflows allocated memory.
- strncpy() is much better, but it always fills the destination by padding the end with as many null characters as necessary. Worse, if it fills the destination then it does not null-terminate. So the next read from that memory will overflow if the correct read size isn't coded.
- dtostrf() has no limit on the number of characters it prints, so if the float has a greater than expected magnitude, the destination will overflow.

A good replacement for the string copy functions would be strlcpy(), but it is not available in all C++ libraries. I can compile it and put it on the Arduino device, but when I upload my code to GitHub, the tests fail because its compiler doesn't recognize strlcpy().

For copying strings, I've added a "strscpy()" function to TC_util.cpp. If the destination is too small to fit the source plus a null terminator, then it will write a truncated, null-terminated string. It will also record in the day's log file that the string was truncated to avoid overflow. There's a similar strscpy_P() function for addresses in PROGMEM.

For printing floats, I've added a "floattostrf()" function to TC_util.cpp. It uses dtostrf() to print the float to a temporary buffer whose size is 10 bytes larger than the destination, and then it uses strscpy() to copy that buffer to the destination. Therefore there will be appropriate logging if the string is truncated. Eventually it might be better to replace "3552148" with "99999" instead of "35521", for example, but I think even truncation is preferable to the current (potential) behavior of overflowing the destination.

I hope this is a positive contribution; I'd be happy to hear feedback!


size.txt
```diff
@@ -1 +1 @@
-Global variables use 2356 bytes (28%) of dynamic memory, leaving 5836 bytes for local variables. Maximum is 8192 bytes.
+Global variables use 2294 bytes (28%) of dynamic memory, leaving 5898 bytes for local variables. Maximum is 8192 bytes.
```
je-foster, line 1: This reduction is from TankController.cpp

src/TC_util.cpp
```diff
@@ -0,0 +1,58 @@
...
+  unsigned long sourceLength = strnlen(source, sizeOfDestination);
+  if (sourceLength < sizeOfDestination) {
+    // Put a null terminator in the final byte
+    *((char *)memcpy(destination, source, sourceLength) + sourceLength) = '\0';
+    return 0;
+  } else {
+    // Put a null terminator in the final byte
+    *((char *)memcpy(destination, source, sizeOfDestination - 1) + sizeOfDestination - 1) = '\0';
+    serial(F("WARNING! String \"%s\" was truncated to \"%s\""), source, destination);
+    // TODO: Log a WARNING that a string was truncated
```
je-foster, line 18: I'd prefer there to be a separate notification process for warnings rather than hoping that someday someone reads the log files carefully. Maybe send e-mails or text messages to interested parties if that were possible.

src/TankController.cpp
```diff
@@ -251,29 +251,37 @@ void TankController::writeDataToSD() {
...
-    dtostrf((float)TempProbe_TC::instance()->getRunningAverage(), 4, 2, currentTemp);
-    dtostrf((float)PHProbe::instance()->getPh(), 5, 3, currentPh);
+    floattostrf((float)TempProbe_TC::instance()->getRunningAverage(), 4, 2, currentTemp, sizeof(currentTemp));
+    floattostrf((float)PHProbe::instance()->getPh(), 5, 3, currentPh, sizeof(currentPh));
   }
-  char targetTemp[10];
-  char targetPh[10];
-  dtostrf(TemperatureControl::instance()->getTargetTemperature(), 4, 2, targetTemp);
-  dtostrf(PHControl::instance()->getTargetPh(), 5, 3, targetPh);
-  static const char header[] = "time,tankid,temp,temp setpoint,pH,pH setpoint,onTime,Kp,Ki,Kd";
```
je-foster, line 261: Moving this string to PROGMEM saves 62 bytes of RAM.

src/Devices/LiquidCrystal_TC.cpp
```diff
@@ -68,6 +68,7 @@ void LiquidCrystal_TC::writeLine(const char* text, uint16_t line) {
 void LiquidCrystal_TC::writeLine(const __FlashStringHelper* text, uint16_t line) {
   // copy from program memory to SRAM and then call other function
   char buffer[17];
-  strcpy_P(buffer, (PGM_P)text);
+  size_t line_length;
```
jgfoster, line 71: Is this variable used?

src/Devices/PHProbe.cpp
```diff
@@ -42,13 +43,13 @@ void PHProbe::clearCalibration() {
 void PHProbe::sendSlopeRequest() {
   // Sending request for Calibration Slope
   Serial1.print(F("SLOPE,?\r"));
-  strncpy_P(slopeResponse, (PGM_P)F("       Slope requested!"), sizeof(slopeResponse));  // Flawfinder: ignore
+  strscpy_P(slopeResponse, (PGM_P)F("       Slope requested!"), sizeof(slopeResponse));  // Flawfinder: ignore
```
jgfoster, line 46: Do we still need `Flawfinder: ignore` on this line (and following ones)? I wonder if it is finding potential bugs with public library functions, but since we have our own functions it doesn't try to find problems. On the other hand, the `_P` version would probably not be known.

src/TC_util.cpp
```diff
@@ -0,0 +1,58 @@
...
+    return 1;
+  }
+}
+
+// In this function, the source is treated as an address in PROGMEM, not RAM
+// Example: strscpy(buffer, source, sizeof(buffer));
```
jgfoster, line 24: Is the example supposed to be showing how to use this function? If so, should it have `_P` on it? It feels like this was just copied from line 7. Is it usually used with something like `(PGM_P)F("...")` as the source?

test/SeeDeviceAddressTest.cpp
```diff
@@ -19,7 +19,7 @@ unittest(testOutput) {
   tc->loop();
   assertEqual("192.168.1.10    ", display->getLines().at(0));
   char buffer[8];
-  strncpy(buffer, display->getLines().at(1).c_str(), 7);  // Flawfinder: ignore
+  strscpy(buffer, display->getLines().at(1).c_str(), sizeof(buffer));  // Flawfinder: ignore
```
jgfoster, line 22: Do we still need Flawfinder?

src/Devices/LiquidCrystal_TC.cpp
je-foster: No, it sure isn't, and I have no memory of that line of code. Thanks for catching that!

src/TC_util.cpp
je-foster: Yes, this was copied and not edited, and your suggested fix is good.

src/Devices/PHProbe.cpp
je-foster: You're right, this flag is no longer needed.

test/SeeDeviceAddressTest.cpp
je-foster: No, apparently not.

jgfoster (approved):

PR #376, opened 2023-03-08
Copy expected value from flash memory to do comparison to buffer

jgfoster:
Otherwise we are comparing SRAM to FLASH and that should not work.


jgfoster:
It appears that the `memcmp_P()` function follows the convention of the other `_P` functions in that the first pointer is to SRAM and the second pointer is to Flash. In this case the code is correct.

PR #378, opened 2023-03-09
Libraries and clearing display

Suocesky:
With the repository merge, some extra libraries were needed and installed. 
Instances of = Tank('', '') have been removed and been replaced with either a clearTank function that sets the current tank to empty, or a bool function checking if the current tank is empty or not.
The display and tank name header on the web page will now clear themselves when the current tank is deleted from the tank list.


extras/webClient/lib/components/app_drawer.dart
```diff
@@ -76,12 +76,13 @@ class AppDrawer extends StatelessWidget {
           Align(
             alignment: Alignment.topRight,
             child: FloatingActionButton(
-                onPressed: () {
-                  appData.removeTank(appData.currentTank);
-                  appData.currentTank = Tank("", "");
-                },
-                tooltip: 'Remove Tank',
-                child: const Icon(Icons.delete_sharp)),
+              onPressed: () {
```
jgfoster, line 79: Is this proper format?

extras/webClient/lib/model/app_data.dart
```diff
@@ -78,16 +84,25 @@ class AppData with ChangeNotifier {
     // ignore result
     await TcInterface.instance.get(tank.ip, 'current');
     _tankList.add(tank);
+    refreshDisplay();
     notifyListeners();
     writeTankList(tankList);
   }
 
   void removeTank(tank) {
     _tankList.remove(tank);
```
jgfoster, line 93: Only clear and refresh if current tank is displayed.

jgfoster (changes requested):

extras/webClient/lib/components/app_drawer.dart
jgfoster: Looks good!

extras/webClient/lib/model/app_data.dart
jgfoster: Looks good!

jgfoster (approved):

prestoncarman (approved):

PR #385, opened 2023-04-28
 Allow longer timeout when building file list

jgfoster:
The timeout was hard-coded at 5 seconds and building a file list takes about 4.5 seconds when it is successful.


prestoncarman (approved):
LGTM

PR #386, opened 2023-05-03
Preparing support for an arbitrary path defined in an uploaded file

Suocesky:
This pull request adds front end functionality for uploading files from the user's computer to storage of the tank controller. The button to do so is at the bottom of the information page, and the function is hidden using the same version check as in the Shortcutsv2 pull request. A tank controller with a version older than the currently hard coded version will instead return a popup advising that the feature is not enabled on their current controller.


prestoncarman:
Can you explain a scenario were this feature would be useful on the Tank Controller?

I am not sure I understand why you would want to have files on the Tank Controller.

je-foster:
I share Preston's question, and I would have some concerns about adding support for PUT/POST to the Tank Controller.

extras/webClient/lib/view/information_page.dart
```diff
@@ -89,20 +158,47 @@ class Information extends StatelessWidget {
...
+                ),
+              ),
+              // This outer ternary is required to prevent the screen flashing an error
+              // for a split second while appData.information loads to do the check.
+              (appData.information['Version'] != null)
+                  ? OutlinedButton(
+                      onPressed: versionCheck(appData.information['Version'])
+                          ? () {
+                              unawaited(
+                                startWebFilePicker(
```
prestoncarman, line 188: Can we limit by file type and size?

extras/webClient/lib/view/information_page.dart
```diff
@@ -89,20 +158,47 @@ class Information extends StatelessWidget {
...
+                                  appData.currentTank.ip.toString(),
+                                ),
+                              );
+                            }
+                          : () {
+                              showNoFileDialog();
+                            },
+                      child: const Text('Add File'),
```
prestoncarman, line 196: Change to something like "Upload arbitrary path CSV file"

Suocesky: Done

extras/webClient/lib/view/information_page.dart
Suocesky: I have added these in. Currently it is limited to file types of .rtf, .txt, and .csv that are less than 10KB in size, though this can be changed if needed.

jgfoster (approved):

Suocesky:
@all-contributors please add @Suocesky for Code

PR #391, opened 2023-06-07
Ethan Ford Pull

ethanfordwork:
Requesting a Pull.

### DIFF
--- a/src/Devices/EthernetServer_TC.cpp
+++ b/src/Devices/EthernetServer_TC.cpp
@@ -52,9 +52,19 @@ void EthernetServer_TC::echo() {
   } else {
     buffer[i - 3] = '\0';
     serial(F("echo \"%s\""), buffer + 19);
-    sendHeadersWithSize(strnlen(buffer, sizeof(buffer)) - 19);
-    client.write(buffer + 19);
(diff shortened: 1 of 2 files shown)


jgfoster:
Why three new submodules and a file discussing BitCoin security?

src/Devices/EthernetServer_TC.cpp
```diff
@@ -52,9 +52,19 @@ void EthernetServer_TC::echo() {
...
+    String timestamp = now.toString();
+
+    // the response
+    JSONBuilder builder;
+    builder.addProperty("timestamp", timestamp);
+    builder.endObject();
+    String reponse = builder.toString();
+
+    sendHeadersWithSize(response.length());
+    client.write(response.c_str()) : state = FINISHED;
```
jgfoster, line 67: What is intended by the ':' character? Is that supposed to be a ';'?

jgfoster:
The `Date` field should be part of the HTTP headers, not part of the body. This would require a change to `EthernetServer_TC::sendHeadersWithSize()` and similar methods. We would also need tests that show proper functionality.

jgfoster:
As we discussed, this can be started over.

PR #393, opened 2023-06-08
Refactor endpoints

jgfoster:
* Use `/data` rather than `/current`
* Rename `Information` to `Current Data` in web UI


prestoncarman (approved):
LGTM after you fix the version number

PR #397, opened 2023-06-13
Use HTTP instead of HTTPS for Flutter app

### LINKED ISSUE #362: Mixed Content Blocking Problems
In Chrome and Firefox the website doesn't show data from the device due to "Mixed Content Blocking." That is, while the primary site (GitHub.io) is https, the device itself is only http. This means that the lock that shows in the address bar is no longer safe. To fix this we need to do one of the following:

1. Disable mixed content blocking in the browser (workarounds are available for [Chrome](https://support.google.com/chrome/answer/99020?hl=en&visit_id=637835792046678344-3519894183&rd=1) and [Firefox](https://support.mozilla.org/en-US/kb/mixed-content-blocking-firefox));
2. Add https support to the device (though this seems [doubtful](https://forum.arduino.cc/t/is-it-possible-to-apply-ssl-to-arduino-mega/445363));
3. Serve the primary site using http from another server (that is not an option for [GitHub Pages](https://docs.github.com/en/pages/getting-started-with-github-pages/about-github-pages#limits-on-use-of-github-pages)); and,
4. Provide a new primary site that acts as an intermediary between the user and the device. We could do this by writing our own application or by just using something like Nginx to act as a reverse proxy to take https in and proxy to the devices using http. To support this we should probably also ask for a port number for each device (defaulting to 80).

jgfoster:
This should avoid the mixed content blocking problems (fix #362).

### DIFF
--- a/src/Devices/EthernetServer_TC.cpp
+++ b/src/Devices/EthernetServer_TC.cpp
@@ -502,7 +502,7 @@ void EthernetServer_TC::sendDisplayRedirect() {
 void EthernetServer_TC::sendHomeRedirect() {
   const __FlashStringHelper *response_303 =
       F("HTTP/1.1 303 See Other\r\n"
-        "Location: https://open-acidification.github.io/TankControllerManager/\r\n"
+        "Location: http://oap.cs.wallawalla.edu/\r\n"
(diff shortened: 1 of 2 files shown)


je-foster (approved):
This looks good to me.

PR #407, opened 2023-07-13
Update target before updating LCD and log

jgfoster:
The previous code updated the LCD and log before recalculating the target pH and temperature. This made the tests somewhat confusing since changes would not be seen till loop() was called a second time. Changing the order has caused some timing changes in the tests, but that is to be expected. Also, I discovered that the LCD setCursor() and print() can take 1-3 ms so I've optimized it so they skip the calls if there is no change.


prestoncarman (approved):
LGTM

Issue #325, opened 2022-08-12
pH Calibration options for 1, 2, or 3 points


jgfoster:
At the start of the pH Calibration, ask whether the user wants a 1-point, 2-point, or 3-point calibration. As it is, accepting 0.0 for low and high creates a very strange calibration.

jgfoster:
- Don't show old calibration settings.
- Verify that calibration happens at all.

jgfoster:
Fixed by #415.

PR #415, opened 2023-07-18
pH Probe Calibration

je-foster:
Addresses #325. When the user is calibrating the pH probe:

- First the user is asked whether there will be 1, 2, or 3 calibration points.
- After the calibration points are submitted, a query is sent to the probe asking how many calibration points are active, and the probe's response is displayed to the user. This must be dismissed by the user pressing a key--it does not time out.
- The device remains in "calibration mode" for the duration of this process until the user dismisses the response. (Previously the device was not in calibration mode during the waits between entering the midpoint and the lowpoint and between entering the lowpoint and the highpoint.)

### DIFF
--- a/src/Devices/PHProbe.cpp
+++ b/src/Devices/PHProbe.cpp
@@ -43,30 +43,30 @@ void PHProbe::clearCalibration() {
 void PHProbe::sendCalibrationRequest() {
   // Sending request for calibration status
   Serial1.print(F("Cal,?\r"));
-  strscpy_P(probeResponse, F("     Calib requested!"), sizeof(probeResponse));
+  strscpy_P(calibrationResponse, F("R"), sizeof(calibrationResponse));
 }
 
 void PHProbe::getCalibration(char *buffer, int size) {
   // for example "?Cal,2" or "     Calib requested!"
-  if (strnlen(probeResponse, sizeof(probeResponse)) > 5) {  // Flawfinder: ignore
-    strscpy(buffer, probeResponse + 5, size);               // Flawfinder: ignore
+  if (strcmp_P(calibrationResponse, PSTR("R"))) {
+    strscpy_P(buffer, F("Calib requested!"), sizeof(buffer));
   } else {
-    buffer[0] = '\0';
+    strscpy(buffer, calibrationResponse, size);
   }
 }
 
 void PHProbe::sendSlopeRequest() {
   // Sending request for Calibration Slope
   Serial1.print(F("SLOPE,?\r"));
-  strscpy_P(probeResponse, F("       Slope requested!"), sizeof(probeResponse));
+  strscpy_P(slopeResponse, F("R"), sizeof(slopeResponse));
 }
 
 void PHProbe::getSlope(char *buffer, int size) {
-  // for example "?SLOPE,99.7,100.3, -0.89" or "       Slope requested!"
-  if (strnlen(probeResponse, sizeof(probeResponse)) > 10) {  // Flawfinder: ignore
-    strscpy(buffer, probeResponse + 7, size);                // Flawfinder: ignore
+  // for example "99.7,100.3, -0.89" or "R"
+  if (strcmp_P(slopeResponse, PSTR("R"))) {
+    strscpy_P(buffer, F("Slope requested!"), sizeof(buffer));
   } else {
-    buffer[0] = '\0';
+    strscpy(buffer, slopeResponse, size);
   }
 }
 
@@ -89,10 +89,10 @@ void PHProbe::serialEvent1() {
         serial(F("PHProbe serialEvent1: \"%s\""), string.c_str());
         if (string.length() > 7 && string.substring(0, 7) == "?SLOPE,") {
           // for example "?SLOPE,16.1,100.0"
-          strscpy(probeResponse, string.c_str(), sizeof(probeResponse));  // Flawfinder: ignore
+          strscpy(slopeResponse, string.c_str() + 7, sizeof(slopeResponse));  // Flawfinder: ignore
         } else if (string.length() > 5 && string.substring(0, 5) == "?Cal,") {
           // for example "?Cal,2"
-          strscpt(probeResponse, string.c_str(), sizeof(probeResponse));  // Flawfinder: ignore
+          strscpy(calibrationResponse, string.c_str() + 5, sizeof(calibrationResponse));  // Flawfinder: ignore
         }
       }
     }
--- a/src/Devices/PHProbe.cpp
+++ b/src/Devices/PHProbe.cpp
@@ -74,7 +74,7 @@ void PHProbe::sendSlopeRequest() {
   Serial1.print(F("SLOPE,?\r"));
 #if defined(ARDUINO_CI_COMPILATION_MOCKS)
   GodmodeState *state = GODMODE();
-  char buffer[10];
+  char buffer[25];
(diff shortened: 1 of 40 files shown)


src/Devices/PHProbe.cpp
```diff
@@ -100,27 +136,36 @@ void PHProbe::setHighpointCalibration(float highpoint) {
...
   char buffer[16];
   snprintf_P(buffer, sizeof(buffer), (PGM_P)F("Cal,mid,%i.%03i\r"), (int)midpoint, (int)(midpoint * 1000 + 0.5) % 1000);
   Serial1.print(buffer);  // send that string to the Atlas Scientific product
+#if defined(ARDUINO_CI_COMPILATION_MOCKS)
+  this->calibrationPoints = 1;
+#endif
   serial(F("PHProbe::setMidpointCalibration(%i.%03i)"), (int)midpoint, (int)(midpoint * 1000) % 1000);
 }
 
 #if defined(ARDUINO_CI_COMPILATION_MOCKS)
```
jgfoster, line 165: I see that you are setting the calibrationPoints earlier; I don't have the ability to search for senders so don't know how this method is used.

src/UIState/NumberCollectorState.h
```diff
@@ -84,4 +87,13 @@ class TestIntNumCollectorState : public TestNumCollectorState {
   }
 };
 
+class TestNoCurrentNumCollectorState : public TestNumCollectorState {
```
jgfoster, line 90: When I read the name of this class I thought that there was no number collector; perhaps the class name could be `TestNumCollectorStateWithNoCurrent` (but it didn't take much to figure out what is intended).

src/UIState/Wait.cpp
```diff
@@ -3,6 +3,13 @@
 #include "MainMenu.h"
 #include "TC_util.h"
 
+/**
```
jgfoster, line 6: Nice documentation!

src/UIState/Wait.cpp
```diff
@@ -14,8 +21,27 @@ Wait::Wait(TankController *tc, uint16_t msDelay, UIState *nextState) : UIState(t
...
+ */
+Wait::~Wait() {
+  if (nextState) {
+    delete nextState;
+    nextState = nullptr;
+  }
+}
+
+bool Wait::isInCalibration() {
+  if (nextState) {
```
jgfoster, line 36: This could be "simplified" (or at least fewer lines):
return nextState ? nextState->isInCalibration() : false;

jgfoster (changes requested):

src/Devices/PHProbe.cpp
```diff
@@ -38,21 +44,47 @@ PHProbe::PHProbe() {
...
+  snprintf_P(buffer, sizeof(buffer), (PGM_P)F("?CAL,%i\r"), this->calibrationPoints);
+  GODMODE()->serialPort[1].dataIn = buffer;  // the queue of data waiting to be read
+#endif
+  strscpy_P(calibrationResponse, F("Requesting..."), sizeof(calibrationResponse));
+}
+
+void PHProbe::getCalibration(char *buffer, int size) {
+  // for example "2" or "Requesting..."
+  strscpy(buffer, calibrationResponse, size);
+#if defined(ARDUINO_CI_COMPILATION_MOCKS)
```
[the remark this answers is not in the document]
je-foster: As written:

1. sendCalibrationRequest() populates the serial port but does not simulate the serial event.
2. getCalibration() returns the placeholder "Requesting..." string, simulating that the EZO pH probe has not yet replied the the request, but then triggers the serial event of receiving the message causing it to load the response into calibrationResponse.
3. A second getCalibration() returns the simulated response...and then triggers another serial event??

I don't like this anymore. I'll remove this mock part and trigger serialEvent1 in the test instead.

src/Devices/PHProbe.cpp
```diff
@@ -6,6 +6,12 @@
 #include "Devices/Serial_TC.h"
 #include "TC_util.h"
 
+#if defined(ARDUINO_CI_COMPILATION_MOCKS)
```
jgfoster, line 9: Not needed here any more?

src/Devices/PHProbe.cpp
```diff
@@ -72,9 +85,13 @@ void PHProbe::serialEvent1() {
         value = string.toFloat();
       } else if (string[0] == '?') {  // answer to a previous query
         serial(F("PHProbe serialEvent1: \"%s\""), string.c_str());
-        if (string.length() > 7 && string.substring(0, 7) == "?SLOPE,") {
+        if (string.length() > 7 && memcmp_P(string.c_str(), F("?SLOPE,"), 7) == 0) {
           // for example "?SLOPE,16.1,100.0"
-          strscpy(slopeResponse, string.c_str(), sizeof(slopeResponse));  // Flawfinder: ignore
+          strscpy(slopeResponse, string.c_str() + 7, sizeof(slopeResponse));  // Flawfinder: ignore
```
jgfoster, line 90: Try without Flawfinder to see if it helps (and on line 94).

test/PHProbeTest.cpp
```diff
@@ -21,14 +21,52 @@ unittest(serialEvent1) {
...
   assertEqual(7.125, pPHProbe->getPh());
+  assertEqual("99.7,100.3,-0.89", pPHProbe->getSlopeResponse());
+}
+
+unittest(clearCalibration) {
+  GodmodeState *state = GODMODE();
+  state->reset();
+  assertEqual("", state->serialPort[1].dataOut);
+  PHProbe::instance()->clearCalibration();
+  assertEqual("Cal,clear\r", state->serialPort[1].dataOut);
```
jgfoster, line 43: Case?

test/PHProbeTest.cpp
```diff
@@ -21,14 +21,52 @@ unittest(serialEvent1) {
...
+}
+
+unittest(sendCalibrationRequest) {
+  GodmodeState *state = GODMODE();
+  state->reset();
+  assertEqual("", state->serialPort[1].dataOut);
+  PHProbe::instance()->sendCalibrationRequest();
+  assertEqual("CAL,?\r", state->serialPort[1].dataOut);
+  char buffer[20];
+  PHProbe::instance()->getCalibration(buffer, sizeof(buffer));
```
jgfoster, line 53: Need buffer?

test/PHProbeTest.cpp
```diff
@@ -82,9 +120,12 @@ unittest(setHighpointCalibration) {
 unittest(sendSlopeRequest) {
   GodmodeState *state = GODMODE();
   state->reset();
-  state->serialPort[0].dataOut = "";
+  assertEqual("", state->serialPort[1].dataOut);
   PHProbe::instance()->sendSlopeRequest();
-  assertEqual("SLOPE,?\r", GODMODE()->serialPort[1].dataOut);
+  assertEqual("SLOPE,?\r", state->serialPort[1].dataOut);
+  char buffer[20];
+  PHProbe::instance()->getSlope(buffer, sizeof(buffer));
```
jgfoster, line 127: buffer?

jgfoster (changes requested):

src/UIState/Wait.cpp
je-foster: Well I typed "/**" and VSCode did the rest...

jgfoster (approved):

jgfoster:
Fix #325.

PR #417, opened 2023-07-18
Remove `static_cast<>()` from `setNextState()`

jgfoster:
While technically correct, it is not necessary to cast a subclass as a superclass when that is what the method expects.


je-foster (approved):
Looks like an improvement to me!

Issue #132, opened 2021-05-05
Include Git SHA in version


jgfoster:
```
echo "#define VERSION \""`git describe --abbrev=4 --dirty --always --tags`"\"" > Version.h
```

prestoncarman:
Stack overflow has a few options: 
<https://stackoverflow.com/questions/1704907/how-can-i-get-my-c-code-to-automatically-print-out-its-git-version-hash>

jgfoster:
The challenge here is that for the value to in the program it must be in a .cpp or .h file. But since we want to reflect the current Git state, any edits to source code will make the Git checkout dirty. One solution is to regenerate the file on each compile, and the Arduino IDE has hooks for this, but you need to edit the toolchain (but see [this](https://github.com/arduino/Arduino/issues/11663)).

PR #418, opened 2023-07-19
Use Git commit hooks to get version

### LINKED ISSUE #410: Show Flutter project version
When running the Flutter application we should show the version so that users can tell if they have the most recent code.

jgfoster:
This adds pre-commit and post-commit scripts to the scripts directory. These scripts are copied to `.git/hooks` by the test.sh and testAndBuild.sh scripts where they are run automatically. 

Before a commit the code uses `git describe --tags` to get a version description. This version description is then placed in two files, `src/Version.h` and `extras/web_client/lib/model/version.dart`. The code then uses these files as the version to display on the device and in the web client (at the bottom of the drawer).

We have a self-referential problem: To put the version in a file that is part of version control, it must be edited before a commit. But the description give by Git will be of the status before the commit (or a description of the _prior_ commit) and we really want to have the version reflect current commit. To make this work you need to commit and tag then do another commit. That way the description will be of the tag alone. 

But then there could be other changes! To address this problem we look at the modified files and if there are changes to anything other than the respective version files then we append a '+' to the version description. If this is not present, then the version description is accurate. 

As a general practice, one should do the initial tag locally (not pushed) and then retag the second commit and move the tag. Then push the second commit and use it for the release.

Fix #410.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -77,19 +77,19 @@ $(BIN)/EEPROMTest.cpp.bin: $(BIN)/libarduino.so $(TEST)/EEPROMTest.cpp
 $(BIN)/EnablePIDTest.cpp.bin: $(BIN)/libarduino.so $(TEST)/EnablePIDTest.cpp
 	$(GPP_TEST) -o $(BIN)/EnablePIDTest.cpp.bin $(TEST)/EnablePIDTest.cpp -larduino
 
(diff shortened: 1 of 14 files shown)


jgfoster:
Fix #132 (though that issue has a slightly different `git describe` command.

Issue #420, opened 2023-07-20
AppData.instance type issues


jgfoster:
The getter for `AppData.instance` is tied to a variable that is initially null so the type is simply Object. We should change this to a method that is known to return an instance of AppData so that `var` will correctly infer the type. This will show us "Asynchronous function invoked in a non-'async' function" as problems.

jgfoster:
The same issue exists for `TcInterface.instance`. This identifies a few more "discarded_futures" isses.

PR #421, opened 2023-07-20
Ensure type is known for AppData and TcInterface singletons

jgfoster:
We were using a property getter for a singleton that starts out as null, so the type was simply Object. By switching to a method getter we can specify the type as not null and this provides better type information (and highlights several "discarded futures" problems that already exist and should be addressed in the future).


jgfoster:
Fix #420.

PR #425, opened 2023-07-22
pH Calibration interface changes

je-foster:
This renames "pH-midpoint", "pH-lowpoint", and "pH-highpoint" depending on the calibration method:
- 1-point calibration: "Buffer pH"
- 2-point calibration: "Higher buffer pH" and "Lower buffer pH"
- 3-point calibration: "Mid buffer pH", "High buffer pH", and "Low buffer pH" (the last two have been reordered)

After calibration, the confirmation message now times out after 60 seconds.


jgfoster:
part of a later PR that has been merged

Issue #403, opened 2023-07-11
Query slope on startup and after pH calibration


jgfoster:
- Warn if pH slope is out of range (initially hard-coded at 95-105%).
- Show on LCD with rapid blinking pH

jgfoster:
* We will add two new boolean flags to the system:
  * A "bad calibration" boolean with the probe instance (volatile); and
  * An "ignore bad calibration" boolean in the EEPROM (nonvolatile).
* On a pH calibration, we will _synchronously_ clear the "ignore bad calibration" flag.
* On startup and after a pH calibration, we will send an _asynchronous_ slope request to the device. 
* On receipt of a slope response, we will save the string and set the "bad calibration" flag based on the response.
* The main menu loop will rapidly blink if "bad calibration" is true.
* The main menu loop will check both flags and if "bad calibration" is true and "ignore bad calibration" is false, then the device will switch to a "bad calibration" UI state (which is "in calibration" so disable the bubbler) and present a screen that asks the user whether to ignore the bad calibration (which sets the nonvolatile flag) or start a new calibration.

je-foster:
I think that the EEPROM default value is 0xffffffff which represents the float -1 which converts to the boolean true. So instead of an "ignore bad calibration" boolean in EEPROM, I suggest a "warn for bad calibration" boolean. Then the default behavior is to warn about an out-of-range calibration.

je-foster:
This issue is resolved by #427.

PR #427, opened 2023-08-01
Check for out-of-range pH calibration slopes

je-foster:
- On boot and with every pH recalibration, we send a SLOPE? request to the pH probe.
- Whenever the pH probe reports a slope less than 95% of ideal or more than 105% of ideal, this is flagged as a "bad calibration."
- In the SeePHCalibration state (which follows every new calibration), a "bad calibration" flag causes an immediate transition to a BadPHCalibration state.
- In the BadPHCalibration state, the user is shown the pH slopes with a warning that this looks like a bad calibration. This is an "in calibration" state that does not time out. The user must choose whether to clear/reset the calibration or to accept the bad calibration. In the latter case, an "ignore bad calibration" flag is set on the nonvolatile EEPROM.
- In the MainMenu state, a "bad calibration" flag with an "ignore bad calibration" flag causes the characters "pH" to blink with the "=" on the display. A "bad calibration" flag without an "ignore bad calibration" flag causes an immediate transition to a BadPHCalibration state. 
- Whenever the pH probe reports a slope between 95% and 105% of ideal, both the "bad calibration" flag and the "ignore bad calibration" flag are reset.
- Whenever the user clears the calibration or sets a new calibration point, both the "bad calibration" flag and the "ignore bad calibration" flag are reset.

This pull request is intended to resolve issue #403.


src/Devices/PHProbe.cpp
```diff
@@ -3,8 +3,10 @@
 #include <avr/wdt.h>
 #include <stdlib.h>
 
+#include "Devices/EEPROM_TC.h"
 #include "Devices/Serial_TC.h"
 #include "TC_util.h"
+#include "TankController.h"
```
je-foster, line 9: Is TankController.h needed?

jgfoster (approved):

PR #430, opened 2023-08-04
Renaming

je-foster:
Renames many classes, functions, and variables:

- "TempProbe_TC" to "ThermalProbe_TC"
- "SeeTempCalOffset" to "SeeThermalCorrection"
- "setTempSetType" to "setThermalFunctionType"
- etc.

The words "temperature" and "temporary" share the abbreviation "temp", which is ambiguous. I am in favor of reserving "temp" for "temporary" because it is a latin root for "time". I prefer to use "therm" where possible for temperature.


extras/mockUI/libTC.cpp
```diff
@@ -51,11 +51,11 @@ float eeprom(uint8_t index) {
...
       return EEPROM_TC::instance()->getPh();
     case 1:
-      return EEPROM_TC::instance()->getTemp();
+      return EEPROM_TC::instance()->getThermalTarget();
     case 2:
       return EEPROM_TC::instance()->getTankID();
     case 3:
-      return EEPROM_TC::instance()->getCorrectedTemp();
+      return EEPROM_TC::instance()->getThermalCorrection();
     case 4:
```
jgfoster, line 59: Nice. This is more than just "temp" cleanup: `getCorrectedX()` implies something different than `getXCorrection()`. You are also removing other ambiguities.

src/Devices/JSONBuilder.cpp
```diff
@@ -56,31 +56,32 @@ int JSONBuilder::buildCurrentValues() {
   uint16_t minutes = (ms - (days * 86400000) - (hours * 3600000)) / 60000;
   uint16_t seconds = (ms - (days * 86400000) - (hours * 3600000) - (minutes * 60000)) / 1000;
```
jgfoster, line 58: I don't understand how the formatting rules liked one then the other, but at least we are following some automated rules!

jgfoster (approved):
That does look like it was rather tedious. Thanks for the effort. Good communication is important to understanding.

PR #431, opened 2023-08-06
Include path edits for tests

jgfoster:
Remove directory from include file path in tests (they are needed in the library when compiled by the Arduino IDE). They are syntatically legal but somewhat confusing since they aren't always present in the test sources. With this they will be present in the test files but not in the files in the src directory.


je-foster:
"With this they will be present in the test files but not in the files in the src directory." Is it the opposite?

je-foster (approved):

jgfoster:
Yes, I got that backwards!

PR #432, opened 2023-08-07
Refactor Directories

jgfoster:
Prior to this we had two subdirectories, 'Devices' and 'UIState'. The Devices directory initially had wrappers for external libraries that controlled devices (LCD, Keypad, etc.) but as other files were added, it came to have everything that wasn't a UIState. Later DataLogger_TC got its name following the convention of most other files in that directory. This was a bit of a misunderstanding since the *_TC naming convention initially was intended to convey a wrapper for the *_CI classes, which were wrappers for the base devices.

This PR attempts to clarify some of the usage. The UIState classes are still in the UIState subdirectory, but the classes that wrap external libraries are in 'wrappers' and other classes are in 'model' (to distinguish them from UI, but other names should be suggested!).

Otherwise, there isn't anything substantive in this commit (no code changes, just path and name changes).


src/UIState/NumberCollectorState.cpp
je-foster, line 1: Could "State" be removed from this class and file name? Otherwise could we add it to the names of all other UIState subclasses?

je-foster (approved):

src/UIState/NumberCollectorState.cpp
jgfoster: Good idea...

PR #433, opened 2023-08-16
Log server

jgfoster:
Dart Shelf HTTP server to upload and download logs.


extras/log_file_server/bin/server.dart
```diff
@@ -0,0 +1,90 @@
...
+
+import 'package:shelf/shelf.dart';
+import 'package:shelf/shelf_io.dart';
+import 'package:shelf_router/shelf_router.dart';
+
+const rootDir = '/var/opt/oap/logs';
+
+// Configure routes.
+final _router = Router()
+  ..delete('/logs/deleteMe.log', _delete)
```
je-foster, line 12: This path is hard-coded because we don't want this functionality on a non-secure server, but we want it for tests?

extras/log_file_server/bin/server.dart
```diff
@@ -0,0 +1,90 @@
...
+    return Response.badRequest(
+        body: 'Content-Length of $length '
+            'did not match body.length of ${body.length}!');
+  }
+
+  // // get remote address
+  // var connectionInfo =
+  //     req.context['shelf.io.connection_info'] as HttpConnectionInfo;
+  // var remoteAddress = connectionInfo.remoteAddress.address;
+  // print('remoteAddress = "$remoteAddress" (${remoteAddress.runtimeType})');
```
je-foster, line 66: These commented lines are for reading IP addresses?

je-foster (approved):

extras/log_file_server/bin/server.dart
jgfoster: Yes, I figured that if someone uploaded data to this file and someone else deleted it I wouldn't feel bad!

extras/log_file_server/bin/server.dart
jgfoster: Yes. While it is best not to leave in commented-out code I thought I might use the remote IP address in some logs.

PR #447, opened 2024-03-13
Refactoring 1 to prepare for event log

jgfoster:
* Rename ThermalProbe_TCTest.cpp to ThermalProbeTest.cpp
* log_file_client
  * Update lint rules
  * Remove comments and sort pubspec.yaml
  * Fix lint issue
* Remove unused include
* Rename `getIsUsingDHCP()` to `isConnectedToNetwork()`
* Use memory compare function that puts string constants in flash memory
* Check to avoid running off end of buffer
* Include filename in serial logging of connection info
* Rename variables to clarify that they are strings rather than floats
* Simplify range check to use `abs()`

### DIFF
--- a/.github/workflows/device-client.yaml
+++ b/.github/workflows/device-client.yaml
@@ -1,6 +1,8 @@
 ---
 name: Build, test, and release device_client
-on: push
+on: 
+  - push
+  - pull_request
 jobs:
   build:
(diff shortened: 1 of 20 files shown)


je-foster (approved):
Looks good.

prestoncarman (approved):
LGTM

PR #454, opened 2024-03-15
Allow user to specify an event log name

jgfoster:
The event log name will be used later. This PR should follow #447.


prestoncarman (approved):
Just add a test for the constructor and remove commented out code.

PR #459, opened 2024-06-14
Current Data Page Expansion

lydiathiel:
This contains my working and tested changes from Spring Quarter. This includes expanded tests, some visual changes on the CurrentData page, and slightly more specific time variable names.

In addition, this expands variables that are able to be viewed, as well as adding more to the editability list. Key values in calculating new target pH and temperature like RampHours are viewable and editable, but values involved in sine calculations are only visible. However, a backend framework has been put in place to receive PUT requests, though function calls are not yet implemented.


jgfoster:
Replaced by #464.

PR #460, opened 2024-06-14
Sine Editability- Flutter App Current Data Page

lydiathiel:
This draft pull request contains proposed changes to implement both pH and Temperature sine value editability from the Flutter app's current data page. 

> The current version frequently has a bug with values increasing significantly, creating memory out of bounds errors. To prevent this from crashing I have expanded the buffer_size from 10 to 200 in TC_util.cpp to allow for further debugging.
`  char large_buffer[buffer_size + 200];`


jgfoster:
Replaced by #465.

PR #466, opened 2024-06-18
Setup to allow a remote log

jgfoster:
This creates a space for a remote log; we'll add to it in a subsequent PR.


src/model/DataLogger.cpp
```diff
@@ -33,20 +33,23 @@ DataLogger* DataLogger::instance() {
  */
 void DataLogger::loop() {
   unsigned long msNow = millis();
-  if (msNow >= nextSDLogTime) {
-    writeToSD();
-    nextSDLogTime = (msNow / (unsigned long)SD_LOGGING_INTERVAL + 1) * (unsigned long)SD_LOGGING_INTERVAL;
+  if (msNow >= nextDataLogTime) {
+    writeToDataLog();
+    nextDataLogTime = (msNow / DATA_LOGGING_INTERVAL + 1) * DATA_LOGGING_INTERVAL;
```
je-foster, line 38: Oh, you're removing the typecasting here and putting it in the constant declaration in the header file. That's good.

src/model/DataLogger.h
```diff
@@ -1,9 +1,29 @@
...
+ * The other events are based on either the PHProbe reporting a new slope or the
+ * EEPROM having a new value written to it. In each case, it means a configuration
+ * changed.
+ *
+ */
+
+// Logging intervals (1 sec, 1 min, 1 min)
+const unsigned long DATA_LOGGING_INTERVAL = 1000;     // 1 sec logging of basic data to SD card
+const unsigned long REMOTE_LOGGING_INTERVAL = 60000;  // 1 min logging of data and other events
+const unsigned long SERIAL_LOGGING_INTERVAL = 60000;  // 1 min logging of data and debugging info
```
je-foster, line 26: What motivates your choice to put remote logging on a timer rather than to have it flagged whenever data is written to the log file? Wouldn't we want to upload notifications of EEPROM changes or extreme thermal/pH probe readings without delay?

je-foster (changes requested):
I'd like to understand the motivation for using a remote-logging time interval rather than having remote logging be triggered by changes to the local log file.

jgfoster:
While I've named the new feature "remote log" I plan to incorporate the same design. The timer deals with how often the Logger writes to the SD card to a file with the name selected by the user. The process of uploading the local "remote" file to the remote "remote" file will be a separate pull request. I'm just breaking it into steps where we have tests passing with each incerement.

je-foster (approved):
Ok, I see. Thanks!

PR #467, opened 2024-06-20
Save some data to a local "remote" log

jgfoster:


src/wrappers/EEPROM_TC.cpp
```diff
@@ -31,11 +33,19 @@ float EEPROM_TC::eepromReadFloat(uint16_t address) {
...
 }
 
+/**
+ * @brief writes the value to the address and triggers an alert in DataLogger
+ *
+ * @param address
+ * @param value
+ */
 void EEPROM_TC::eepromWriteFloat(uint16_t address, float value) {
-  assert(address >= 0);
```
je-foster, line 35: Did this assertion do anything?

src/wrappers/EEPROM_TC.cpp
```diff
@@ -49,11 +59,20 @@ int32_t EEPROM_TC::eepromReadInt(uint16_t address) {
...
 }
 
+/**
+ * @brief writes the value to the address and triggers an alert in DataLogger
+ *
+ * @param address
+ * @param value
+ */
 void EEPROM_TC::eepromWriteInt(uint16_t address, int32_t value) {
   assert(address >= 0);
```
je-foster, line 69: Here the assertion is not removed.

jgfoster: Not for an unsigned number! I'll change it to `!= 0`.

src/wrappers/EEPROM_TC.cpp
jgfoster: Not for an unsigned number! I'll change it to `!= 0`.

src/wrappers/SD_TC.cpp
```diff
@@ -245,3 +254,40 @@ void SD_TC::todaysDataFileName(char* path, int size) {
...
+    DataLogger::instance()->putRemoteFileHeader(buffer, sizeof(buffer), 1);
+    appendDataToPath(buffer, remoteLogName, false);
+    DataLogger::instance()->putRemoteFileHeader(buffer, sizeof(buffer), 2);
+    appendDataToPath(buffer, remoteLogName, false);
+    DataLogger::instance()->putRemoteFileHeader(buffer, sizeof(buffer), 3);
+    appendDataToPath(buffer, remoteLogName);
+  }
+  appendDataToPath(line, remoteLogName);
+  updateRemoteFileSize();
+  // AlertPusher::instance()->pushSoon();
```
je-foster, line 292: So this commented-out line is a placeholder for the next PR, right? It triggers the network process for transferring data from the local "remote" file to its copy on the OAP server, right?

src/wrappers/EEPROM_TC.cpp
je-foster: So the typecasting made the ">= 0" assertion meaningless, I guess?
Regarding "!= 0" though, isn't PH_ADDRESS zero? We want to allow setPH() and getPh() to work, right? So remove these assertions entirely?

src/wrappers/SD_TC.cpp
jgfoster: Yes, exactly...

src/wrappers/EEPROM_TC.cpp
jgfoster: You are right, they should be removed entirely. If we had it to do over, it might have been wise to reserve address zero as unused (location zero in memory is a real location but it is often left unused and treated as an error to catch null pointers when designing operating systems).

je-foster (approved):

PR #468, opened 2024-06-25
Bring in tests from `alert` branch

jgfoster:
Bring in tests for logging and Mean/StdDev.


src/model/PHProbe.cpp
```diff
@@ -124,6 +126,7 @@ void PHProbe::setTemperatureCompensation(float temperature) {
 }
 
 void PHProbe::setHighpointCalibration(float highpoint) {
+  // TODO: the next two lines were commented out in another branch
```
je-foster, line 129: Perhaps it is because those two lines will be executed with the midpoint calibration anyway? I think the midpoint calibration is the one that is always performed whether it is a one-point, two-point, or three-point calibration. Although redundant, it may be ok to leave them here so long as they do not cause extra alerts to the remote log.

src/model/PHProbe.cpp
```diff
@@ -134,6 +137,7 @@ void PHProbe::setHighpointCalibration(float highpoint) {
 }
 
 void PHProbe::setLowpointCalibration(float lowpoint) {
+  // TODO: the next two lines were commented out in another branch
```
je-foster, line 140: same

src/wrappers/ThermalProbe_TC.cpp
```diff
@@ -1,10 +1,22 @@
...
+#endif
+
+#include "model/Statistic.h"
 #include "model/TC_util.h"
 #include "wrappers/DateTime_TC.h"
 #include "wrappers/EEPROM_TC.h"
 #include "wrappers/Serial_TC.h"
 
+//  local statistics
+Statistic myStats;
```
je-foster, line 18: Could we name this something like "thermalSample" or "uncorrectedThermalSample"? And perhaps put it in the header file?

src/wrappers/ThermalProbe_TC.cpp
```diff
@@ -127,11 +137,7 @@ void ThermalProbe_TC::setCorrection(float value) {
  */
 float ThermalProbe_TC::getSampleMean() {
   // This is not used for the Liquid Crystal display, so we don't restrict range to 0 to 99.99
-  if (sampleSize > 0) {
-    return (sumOfSampleValues / sampleSize) + correction;
-  } else {
-    return 0.0;
-  }
+  return myStats.average();
```
je-foster, line 140: I think this should be changed to `return myStats.average() + correction;`

src/wrappers/SD_TC.h
```diff
@@ -22,25 +22,28 @@ class SD_TC {
...
 #if defined(ARDUINO_CI_COMPILATION_MOCKS)
-  char mostRecentHeader[128] = "";
-  char mostRecentLine[128] = "";
-  char mostRecentRemoteEntry[256] = "";
+  char mostRecentDataLogHeader[128] = "";
+  char mostRecentDataLogLine[128] = "";
+  char mostRecentRemoteLogEntry[256] = "";
   void updateAlertFileSizeForTest() {
     updateRemoteFileSize();
   }
```
je-foster, line 49: This method appears to be called only by the commented-out lines in SDTest.cpp. Can it be removed and replaced by updateRemoteFileSize()?

je-foster: Speaking of which, is remoteFileSize used anywhere other than tests?

extras/device_client/web/index.html
je-foster, line 1: What's this about?

src/model/Statistic.h
je-foster, line 1: We looked at the unit test for this yesterday, where the mean and standard deviation of the first ~100 positive integers were calculated. And that test looked good, but I wish there were another one or two tests for the extreme case when all sampled values are equal (no variance). For example, find the mean and standard deviations of the numbers 10.2, 10.2, 10.2, 10.2, 10.2. In this case the standard deviation is the square root of zero. The radicand (zero) is a difference of two values. In my previous testing of a case like this I found that the radicand, which would always be nonnegative if calculated to infinite precision, could be slightly negative due to rounding errors which caused an error when its square root was calculated. Therefore in the alert branch you can see that I passed max(0.0, ...) into the square root function. I don't see that precaution being taken here in lines 241 and 249.

je-foster (changes requested):
I'd like to see at least that change in ThermalProbe_TC.cpp where the temperature correction is applied to the sample mean before it is written in the remote log file.

test/SDTest.cpp
```diff
@@ -182,11 +188,74 @@ unittest(removeFile) {
...
+//   // write data
+//   sd->setRemoteLogName("Tank1");
+//   // sd->writeAlert("line 1");
+//   int size = sd->getRemoteFileSize();
+//   // sd->writeAlert("and 2\nline 3");
+//   char buffer[20];
+//   // get remaining alerts
+//   sd->getAlert(buffer, sizeof(buffer), size);
+//   assertEqual("and 2\nline 3\n", buffer);
+// }
```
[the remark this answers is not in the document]
jgfoster: They have been removed. Other tests will be incorporated or written to work with the new functionality as it is added.

src/model/PHProbe.cpp
jgfoster: Agreed. I'll also note that the midpoint might not be reached if the user cancels after a high or low without completing the midpoint.

src/model/PHProbe.cpp
jgfoster: Comment removed.

src/wrappers/ThermalProbe_TC.cpp
jgfoster: Renamed and moved to header.

src/wrappers/ThermalProbe_TC.cpp
jgfoster: Done.

extras/device_client/web/index.html
jgfoster: This is the `index.html` page served as the beginning web page. It is generated at the beginning of the project by `flutter create` and gets only minor changes (like adding a `favicon.png`). I found that Flutter has changed since we first got the file so I let Flutter recreate it to match the new style. Mostly this moved some JavaScript from the `<body>` of the HTML to `flutter_bootstrap.js`.

jgfoster:
I'm back to looking at things and have made updates and added comments.

je-foster (approved):

PR #473, opened 2024-10-30
Proxy server and HttpClient

anna-st-40:
Implemented proxy server system for development and combined CsvReader and LogListReader into one HttpClient.


jgfoster (approved):

jgfoster:
This is very good. We might explore some other ways to approach the prod/test alternatives, but that can be another time.

PR #476, opened 2024-12-27
Refactor main

jgfoster:


src/model/JSONBuilder.cpp
je-foster, line 1: Please revert these changes.

test/SerialTest.cpp
je-foster, line 1: Change to match new loop timers.

src/TankController.cpp
```diff
@@ -158,25 +161,31 @@ void TankController::handleUI() {
...
  * This is one of two public instance functions.
  * It is called repeatedly while the board is on.
- * (It appears to be called about once every 15 ms.)
  */
 void TankController::loop(bool report_loop_delay) {
   static unsigned long lastTime = 0;
   unsigned long thisTime = millis();
   if (report_loop_delay && lastTime && thisTime - lastTime > 500) {
-    // report unusual delay
     serial(F("unexpected delay of %i ms"), thisTime - lastTime);
```
je-foster, line 169: I think it would be clearer to measure the difference between `currentLoopStart` and `previousLoopStop`

je-foster: This would also be a good place to measure and report the difference between `currentLoopStart` and `previousLoopStart`

src/TankController.cpp
```diff
@@ -158,25 +161,31 @@ void TankController::handleUI() {
...
+  Ethernet_TC::instance()->loop();        // renew DHCP lease (~0ms)
+  EthernetServer_TC::instance()->loop();  // handle any HTTP requests (~0ms)
+  if (report_loop_delay) {
+    static long int count = 0;
+    unsigned long loopTime = millis() - start;
+    if (++count % 10000 == 1 || loopTime > 200) {  // first time through and periodically thereafter
+      serial(F("TankController::loop() - took %lu ms (at %lu sec uptime)"), loopTime, start / 1000);
+    }
+    lastTime = millis();
+  }
```
je-foster, line 188: Here, just set `previousLoopStop` to `millis()` ("previous" isn't quite right, but it's about to be previous)

je-foster (changes requested):
See comments regarding loop delay timers (and its test) and the sine amplitude being nonzero.

je-foster (approved):
It looks good.

PR #478, opened 2025-01-13
Implemented landing page approach to navigating tanks

anna-st-40:


extras/log_file_client/lib/components/tank_card.dart
jgfoster, line 1: I'm not seeing immediately why this is a StatefulWidget. Is there a call to `setState()`?

jgfoster (changes requested):
Some requested changes and some more general questions/comments. But overall it looks good. I like the fact that the graphing itself is simple calls to libraries!

extras/log_file_client/lib/components/graph_view.dart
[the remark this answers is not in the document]
anna-st-40: Removed. I found the reason why I had them - the linter though that returning a null value from parseLogListFromHTML is possible even though it is not. I refactored the function and now there are no problems.

extras/log_file_client/lib/components/tank_card.dart
anna-st-40: Calling initState is the only way I've figured out to have a default HttpClient, since having it be a traditional default value result in an error "The default value of an optional parameter must be constant."
(This is on lines 26-30, I'm not sure how to reference code in comment replies)

extras/log_file_client/lib/components/project_card.dart
```diff
@@ -0,0 +1,79 @@
...
+  @override
+  Widget build(BuildContext context) {
+    return MouseRegion(
+      cursor: SystemMouseCursors.click,
+      child: GestureDetector(
+        onTap: onTap,
+        child: LayoutBuilder(
+          builder: (BuildContext context, BoxConstraints constraints) {
+            // Access the available width and height of the widget
+            final double cardWidth = constraints.maxWidth * 0.93;
```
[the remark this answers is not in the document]
anna-st-40: This is to get a consistent sizing across different window sizes. The width is used to calculate font sizes, image sizes, and horizontal margins.

jgfoster (approved):

extras/log_file_client/lib/components/tank_card.dart
jgfoster: It looks like `httpClient` is passed in. If you made it `required`, then you would not need to initialize it.

PR #479, opened 2025-01-14
Early January Refactoring

jgfoster:
Move more of the alerts branch into main.

### DIFF
--- a/src/wrappers/SD_TC.cpp
+++ b/src/wrappers/SD_TC.cpp
@@ -128,7 +128,17 @@ bool SD_TC::format() {
 }
 
 void SD_TC::getRemoteLogContents(char* buffer, int size, uint32_t index) {
-  // This function is not called in the current codebase
+  buffer[0] = '\0';
+  File file = open(getRemoteLogName(), O_RDONLY);
+  if (file) {
+    file.seek(index);
+    int remaining = file.available();
+    if (remaining > 0) {
+      int readSize = file.read(buffer, min(size - 1, remaining));
+      buffer[readSize] = '\0';
+    }
+    file.close();
+  }
 }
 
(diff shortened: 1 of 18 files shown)


src/wrappers/SD_TC.cpp
```diff
@@ -125,6 +127,20 @@ bool SD_TC::format() {
...
 
+void SD_TC::getRemoteLogContents(char* buffer, int size, uint32_t index) {
+  buffer[0] = '\0';
+  File file = open(getRemoteLogName(), O_RDONLY);
+  if (file) {
+    file.seek(index);
+    int remaining = file.available();
+    if (remaining > 0) {
+      int readSize = file.read(buffer, min(size - 1, remaining));
+      buffer[readSize] = '\0';
```
je-foster, line 138: So Codacy doesn't like something in line 137? We're trusting callers to pass sizeof(buffer) for the second argument, right? Is there supposed to be a way around that?

Maybe it would be good to replace line 138 with something like the following, in case there is a read error.
```
      if (readSize == min(size - 1, remaining)) {
        buffer[readSize] = '\0';
      } else {
        buffer[0] = '\0';
      }
```

je-foster (changes requested):
I posted a comment near Codacy's complaint.

src/wrappers/SD_TC.cpp
je-foster: In case of an error `readSize` might be `-1` I think.

jgfoster: That seems to have fixed it!

jgfoster: After adding your suggested change we still get the same issue, so I'm going to mark it to be ignored.

PR #484, opened 2025-01-28
Updated logic between log_file_server and log_file_client

anna-st-40:


extras/log_file_client/lib/components/tank_card.dart
```diff
@@ -57,27 +57,27 @@ class TankCard extends StatelessWidget {
     );
   }
 
-  FutureBuilder<TankSnapshot> _graphThumbnailBuilder(
+  Expanded _graphThumbnailBuilder(
     Future<TankSnapshot> tankSnapshot,
```
jgfoster, line 61: Rather than explicitly name the subclass that is returned, I typically just name `Widget` as the return type to simplify future refactorings.

extras/log_file_client/lib/utils/http_client.dart
```diff
@@ -231,52 +239,64 @@ class HttpClientTest extends HttpClient {
   // Fetches data from a hard-coded string (short) or the local Dart file (1440 sample lines) for testing purposes
   HttpClientTest();
 
-  late String testHTML =
-      '<html><body><ul><li><a href="/logs/ProjectA-tank-24.log">/logs/ProjectA-tank-24.log</a></li><li><a href="/logs/ProjectA-tank-70.log">/logs/ProjectA-tank-70.log</a></li><li><a href="/logs/ProjectB-tank-58.log">/logs/ProjectB-tank-58.log</a></li><li><ahref="/logs/index.html">/logs/index.html</a></li></ul></body></html>';
+  late String testHTML = '''
```
jgfoster, line 242: Not needed for this PR, but eventually it would be nice to separate the "test" classes and code from the production code so that these strings don't form a part of the executable.

jgfoster (approved):

PR #485, opened 2025-02-05
UI updates

anna-st-40:
Made some of the changes discussed with Dr. Onthank, including:

- Simplifying & updating tank snapshot graphs
- Adding more visible setpoint lines
- Adding a pH/temperature option selector when viewing graph
- Replacing circular loaders with skeleton loaders 

Also, the snapshot API on the log file server now accepts parameters, with a default of 12 hours.


extras/device_client/lib/view/home_page.dart
```diff
@@ -7,6 +7,7 @@ import 'package:device_client/view/current_data_page.dart';
 import 'package:device_client/view/files_page.dart';
 import 'package:flutter/material.dart';
 import 'package:provider/provider.dart';
+// import 'package:web/web.dart' as html;
```
jgfoster, line 10: If you are going to commit commented-out code, add a TODO: with an explanation of what you plan to do with it.

extras/log_file_client/lib/components/graph_view.dart
```diff
@@ -12,41 +13,82 @@ class GraphView extends StatelessWidget {
   final String filePath;
   final HttpClient httpClient;
 
+  @override
+  State<GraphView> createState() => _GraphViewState();
+}
+
+class _GraphViewState extends State<GraphView> {
+  bool _showPH = true;
```
jgfoster, line 21: This is a good reason to use a StatefulWidget!

extras/log_file_client/lib/pages/home_page.dart
```diff
@@ -4,6 +4,8 @@ import 'package:flutter/material.dart';
 import 'package:log_file_client/components/project_card.dart';
 import 'package:log_file_client/pages/project_page.dart';
 import 'package:log_file_client/utils/http_client.dart';
+import 'package:skeletonizer/skeletonizer.dart';
+// import 'package:web/web.dart' as html;
```
jgfoster, line 8: More commented-out code.

extras/log_file_client/lib/utils/http_client.dart
```diff
@@ -95,7 +104,8 @@ abstract class HttpClient {
 
   Future<List<Project>> getProjectList() async {
     // Fetch data from server
-    final data = await fetchData('logs');
+    // final data = await fetchData('logs');
```
jgfoster, line 107: As I recall, we needed to give a full path because in development the reverse proxy sends us to the Flutter server and it does files but not directories. Is that right?

extras/log_file_client/logs/index.html
```diff
@@ -1,10 +1,18 @@
...
+    <hr>
+    <pre>
+        <a href="ProjectA-tank-24.log">ProjectA-tank-24.log</a>
+        <a href="ProjectA-tank-70.log">ProjectA-tank-70.log</a>
+        <a href="ProjectB-tank-58.log">ProjectB-tank-58.log</a>
+    </pre>
+    <hr>
+</body>
+
 </html>
```
jgfoster, line 18: The red circle with a line indicates that there is not a newline at the end of the file. We try to include this.

jgfoster (changes requested):
A couple minor requests...

extras/log_file_client/lib/utils/http_client.dart
anna-st-40: That's right. Adding index.html to all paths ending with `/` caused issues, but I did come up with a very similar solution in which all requests to just `logs` are redirected to 'logs/index.html` by the reverse proxy server.

jgfoster (approved):
Very nice. Thanks!

PR #488, opened 2025-02-08
Remove most testing of serial output

jgfoster:
We (still) have tests to see if the serial commands work, but most of the remaining testing of serial output duplicated other tests and was very brittle. For example, after turning off the header we check to see if the heater is off and then check to see if there is a message in the log that the heater was turned off. As other serial output was added and removed (primarily for debugging) other tests would fail. Also since serial output often includes a timestamp, any slight change in timing would make the tests fail for completely innocent reasons.


jgfoster:
This ended up with too much. I'll break it up some more.

PR #496, opened 2025-02-25
Remote log name

jgfoster:
The SD card seems to support only 8.3 names so we should not use the provided filename on local storage. Instead of storing it with the SD wrapper, we now store it with the RemoteLogPusher class.

### DIFF
--- a/test/SDTest.cpp
+++ b/test/SDTest.cpp
@@ -190,15 +190,15 @@ unittest(removeFile) {
 
 unittest(writeRemoteLog) {
   SD_TC* sd = SD_TC::instance();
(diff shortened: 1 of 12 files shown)


test/SDTest.cpp
```diff
@@ -226,7 +226,7 @@ unittest(getRemoteLogContents) {
   SD_TC* sd = SD_TC::instance();
 
   // write data
-  sd->setRemoteLogName("Tank1");
+  // sd->setRemoteLogName("Tank1");
```
jgfoster, line 229: commented-out code

PR #510, opened 2026-05-01
Reset EEPROM

### LINKED ISSUE #509: Option to clear/ignore EEPROM
Kirt reports that Lloyd Trueblood has a device that reboots constantly and appears to have bad data in EEPROM. We should have an option to reset/ignore EEPROM on boot, probably with a keypress. We have the option to not connect to Ethernet on boot (`Ethernet_TC.cpp:18-19`) if any key is pressed. We should refine this so that one key deals with EEPROM and another deals with Ethernet.

Image

je-foster:
To resolve #509. These changes allow the user to reset EEPROM to factory default. During the TankController boot sequence the user holds '1' on the keypad. This blocks all EEPROM reading/writing while the UI asks the user whether to reset EEPROM. Then the device reboots.

Handling of keypresses during the boot sequence is moved from the Ethernet_TC constructor to TankController instance creation so that keypresses can be used for purposes other than disabling Ethernet.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -71,6 +71,7 @@ all : \
 	$(BIN)/PIDTest.cpp.bin \
 	$(BIN)/PushingBoxTest.cpp.bin \
 	$(BIN)/RemoteLogPusherTest.cpp.bin \
+	$(BIN)/ResetEEPROMTest.cpp.bin \
 	$(BIN)/ResetPHCalibrationTest.cpp.bin \
 	$(BIN)/ResetThermalCalibrationTest.cpp.bin \
 	$(BIN)/SDTest.cpp.bin \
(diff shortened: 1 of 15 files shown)


je-foster:
Codacy checks are failing because they do not have <Arduino.h> or <ArduinoUnitTests.h> or <avr/wdt.h> in their collection of standard library headers. I intend to disable the Code Pattern that checks for missing included header files.

je-foster:
I haven't had any success disabling the Codacy code pattern for this repository that is flagging those Arduino header files as missing. I think the next step to try is to create a Coding Standard at the Organization level in Codacy, to disable the code pattern(s) there, and to apply that coding standard to the TankController repository. But I am not an organization admin for Open-Acidification in Codacy, so I can't create a Coding Standard. I could try disabling all of Cppcheck for the entire TankController repository...

jgfoster (approved):

